PR #6, opened 2021-02-15
Fix event table and rename files

tahini:


bhufmann (approved):
Thanks for this. The events table is now filled.

bhufmann:
@tahini you can merge this one. Thanks for fixing the issue.

Issue #35, opened 2022-01-14
vscode-trace-extension not working with latest trace-viewer depenencies


bhufmann:
With the recent updates in the trace viewer components and its dependencies to use `BigInt` for timestamps, the serialization of messages between the webviews and vscode extension will fail because the standard json serializer doesn't support type `BigInt`. 

It will fail with the following message:
```
TypeError: Do not know how to serialize a BigInt
/usr/share/code/resources/app/out/bootstrap-fork.js:5
stack trace: TypeError: Do not know how to serialize a BigInt
	at JSON.stringify (<anonymous>)
	at y (/usr/share/code/resources/app/out/vs/workbench/services/extensions/node/extensionHostProcess.js:89:3480)
	at g.postMessage (/usr/share/code/resources/app/out/vs/workbench/services/extensions/node/extensionHostProcess.js:96:73489)
	at TraceViewerPanel.setExperiment (/home/eedbhu/git/vscode-trace-extension/vscode-trace-extension/lib/trace-viewer-panel/trace-viewer-webview-panel.js:125:29)
	at /home/eedbhu/git/vscode-trace-extension/vscode-trace-extension/lib/trace-explorer/trace-tree.js:123:19
	at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

For example the` Experiment` or `Trace` classes contain timestamp fields which are `BigInt` and causing serialization problems and the vscode trace extension to fail.

To fix this, each message send between webviews and the vscode trace extension itself need to be updated that `BigInt` values are serialized properly (e.g. using a string) and deserialized to `BigInt` on the receiving side.

This issue needs to be fixed before upgrading to the latest traceviewer components which contain many valuable improvements.

bhufmann:
Using https://github.com/eclipse-cdt-cloud/tsp-typescript-client/pull/56, the serialization / deserialization of BigInt values can be achieved when sending/receiving signals.

bhufmann:
The extension works now with:
- traceviewer-base: 0.2.0-next.7298b48.0
- traceviewer-react-components: "0.2.0-next.7298b48.0"

Serialization is fixed for experiments and it's not an issue for data provider descriptors. Closing this one and open a ticker for improved and more generic serialization implementation.

PR #54, opened 2022-09-14
Fix bigint support in messages sent between webviews and vscode

bhufmann:
TODO: JSONBig.parse() doesn't create bigint if numbers are small. This has to be addressed when deserializing messages. Re-use the normalizer of the tsp-typescript-client to do that.

The following branch in the tps-typescript-client is introducing an utility class and, if accepted, it can be used to solve the deserialization issue:

https://github.com/eclipse-cdt-cloud/tsp-typescript-client/pull/56

Contributes to fixing #35


MatthewKhouzam (approved):
Nice, the overview is added too.

bhufmann:
@ngondalia this pull request upgrades the vscode-trace-extension to the latest traceviewer code and corresponding trace server. It has some shortcuts to make it work. But I think it's good step forward. Any objections merging it?

ngondalia:
> @ngondalia this pull request upgrades the vscode-trace-extension to the latest traceviewer code and corresponding Trace Compass trace server. It has some shortcuts to make it work. But I think it's good step forward. Any objections me merging it?

We are using a custom trace server, so there are some issues I am running into with these changes. Everything works fine the first time i open the trace, but once i close it and re-open it, there is bunch of weird behavior where the views wont load and evrything disappers.. seems like a problem on our server because it doesn't seem to happen on tracecompass server. Any ideas on what it could be? Do i need to make any changes on server-side for this fix?

bhufmann:
> 

Not sure. But I have a suspicion. We changed the API on how the data samples are requested from the backend. Instead of providing a list of timestamps, now a range with start, end and number of samples is used. This was to avoid send large arrays in fetch methods. See [1] for the change. If your custom trace server is not able to handle that the new parameters it would explain some problems with the visualization. See also [2] for the corresponding Trace Compass server change: 

[1] https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/703
[2] https://git.eclipse.org/r/c/tracecompass.incubator/org.eclipse.tracecompass.incubator/+/191752

bhufmann:
@ngondalia I'm going to merge. The updated version has more features and performance improvements, especially in the timegraphs. I hope make your server work with it is easy to do. Please let me know if you have any questions.

ngondalia:
> @ngondalia I'm going to merge. The updated version has more features and performance improvements, especially in the timegraphs. I hope making your server work with it is easy to do. Please let me know if you have any questions.

Yeah, no problem. I am still looking into it why its not working. I have traced the exception down to onOverviewRemoved() in vscode-trace-viewer-container.tsx. For some reason,' this.state' is undefined when this callback occurs and this causes everything to go in a weird state. Still investigating as to why this is occuring. Are there any changes that needs to be made with custom trace types with the introduction of overview view? Still fairly new to this codebase, so not sure how a lot of the things work.

bhufmann:
> Are there any changes that needs to be made with custom trace types with the introduction of overview view? Still fairly new to this codebase, so not sure how a lot of the things work.

With this PR, the overview view is shown only using the Trace Compass trace server. The data provider is hard-coded to a specific data provider. If this data provider doesn't exist on the back-end (e.g. your back-end) then it is not shown. To support properly more work is needed. I opened a tracker for that #55. It's also not possible to open the overview after closing it (trace needs to re-opened) for that nor configure the default overview data provider.

The idea of the overview view is that it shows the whole trace for a given XY data provider (only XY at the moment) and users can see where to zoom in.

PR #81, opened 2023-01-19
Fix these while starting to use README.md and this repo

marco-miller:
This PR is simply made of four (4) trivial yet distinct commits.


bhufmann (approved):
LGTM. Thanks

PR #84, opened 2023-02-02
Fix refresh of views when switching trace tabs

### LINKED ISSUE #63: Refresh of views not working when switching trace tabs
When multiple traces are open and a user switches between trace tabs, the "Opened Traces" and "Available Views" views need to updated with the corresponding data of the selected experiment.

vscode-failing-sync

Rodrigoplp-work:
The wrong object was being passed to the parser.

Fixes #63.

Signed-off-by: Rodrigo Pinto <rodrigo.pinto@calian.ca>


bhufmann (approved):
That fixes the bug. Thanks for the contribution.

PR #92, opened 2023-02-23
Refresh "VIEWS" and "TRACES" after being minimized and re-opened

### LINKED ISSUE #64: "Available Views" not refreshed after being minimized and opened
When the "Available Views" are minimized and then re-opened, the view is not populated for the current selected experiment.

vscode-available-view-sync-bug

bhufmann:
Make sure that selection of current experiments are persisted and restored in both views by sending and handling tabActivatedSignal and experimentSelectedSignal at relevant places.

Also, fix clearing of views and selection when trace panels are closed.

Fixes #64


Rodrigoplp-work (approved):
Well done.

PR #102, opened 2023-03-15
opened-traces: Remove redundant case code lines

marco-miller:
- Remove the `closeTrace` case lines of code, falling through `deleteTrace`'s which are identical.
- Based on the related SonarLint warning in VS Code.


bhufmann (approved):
Looks good to me! Thanks.

PR #103, opened 2023-03-15
webpack config: Remove obsoleted/extra devtool key

marco-miller:
- Since commit `1287ba25` introduced the currently required `"inline-source-map"` value for that key.
- The other, earlier commit `16fe6129` initialized that key prior, to a previously wanted value.
- This fix is also based on the related SonarLint warning in VS Code.

### DIFF
--- a/vscode-trace-webviews/webpack.config.js
+++ b/vscode-trace-webviews/webpack.config.js
@@ -12,7 +12,6 @@ module.exports = {
     path: path.resolve(__dirname, "../vscode-trace-extension/pack"),
     filename: "[name].js"
   },
-  devtool: "eval-source-map",
   devtool: "inline-source-map",
   resolve: {
     extensions: [".js", ".ts", ".tsx", ".json"]


bhufmann:
I'm not sure what the issue this PR is fixing. Looking at the devtools documentation, both entries are correct.  Please let me know what the issue is?

Also, I don't see any SonarLint warning. Did you mean EsLint? I use ESLint and I don't see any warning.

bhufmann:
> I'm not sure what the issue this PR is fixing. Looking at the devtools documentation, both entries are correct. Please let me know what the issue is?
> 
> Also, I don't see any SonarLint warning. Did you mean EsLint? I use ESLint and I don't see any warning.

Ok. I installed SonarLint as well and get the warning that you see. I think the last one (inline-source-map) take precedence. So, I think your change is fine.

bhufmann (approved):
LGTM. Thanks!

PR #119, opened 2023-03-20
launch.json: Remove disallowed stopOnEntry property

marco-miller:
This property was highlighted as disallowed, by VS Code/Codium locally.


bhufmann (approved):
Thanks!

PR #120, opened 2023-03-23
Render new trace server connection status bar item

### LINKED ISSUE #33: Migrate health check
The health check functionality to check if the trace server is running needs to be migrated. Call heath check functionality before opening a trace.

marco-miller:
Add a contextual (secondary, right-side) VS Code status bar item, that renders the current trace server status. Have the extension context pass this status service to each active web-view provider, for them to update the rendered status upon every related message from the view.

Have `TraceExplorerOpenedTracesViewProvider` pass its status service to the `TraceViewerPanel` upon using it, in turn. Without this, there are instances where the latest server status isn't shown as expected. Keep this undefined in `trace-tree.ts` where that is not required.

This makes the server status listener-based. Add this connection listener through the `TspClientProvider` constructor, thus consistently use the latter across. This means also using it in the `TraceViewerContainer` constructor, instead of `TspClient`'s directly prior.

Properly `stringify` (thus parse) the boolean status value, as passing it straight would break the JSON message otherwise.

Explicitly defer showing the item to only once Trace Viewer gets selected for the first time. Like open trace-explorer views which remain shown if unselecting Trace Viewer, keep showing the status item from there on. The VS Code API, anyway, doesn't hold an event that could selectively hide anything upon solely exiting Trace Viewer.

Prefer a standard VS Code status bar item to the side widget approach of `theia-trace-extension`. This should be more standard for such a connection status display, from a VS Code UI (and UX) perspective. This also allows not having to implement a custom widget that differs from the Trace Viewer panels. -The latter being meant to render real views.

Base this simple design of the status bar item on [1] below. Now, use the warning background for a found server, for such a positive status to be highlighted, thus noticeable. Use the usual red background for a trace server that is not found, highlighting that as an error. No other color can or should be used anyway for status bar items, per [1]. The latter reference also includes [2]'s, used to come up with this change.

[1]https://code.visualstudio.com/api/ux-guidelines/status-bar
[2]https://github.com/microsoft/vscode-extension-samples/blob/main/statusbar-sample/src/extension.ts

Use a short (thus usable) enough `statusBarItem.text`, along with a `tooltip` to complement the trace server status narrative.

Overall, with this approach, defer the introduction of the `inversify` module, used in `theia-trace-extension`. This would be to inject instances such as objects involved in this change and likely other ones; some being potential singletons.

Slightly refactor a few surrounding blank lines while editing some of these files.

Fixes #33.

### DIFF
--- a/vscode-trace-extension/src/trace-explorer/available-views/trace-explorer-available-views-webview-provider.ts
+++ b/vscode-trace-extension/src/trace-explorer/available-views/trace-explorer-available-views-webview-provider.ts
@@ -4,6 +4,7 @@ import { Experiment } from 'tsp-typescript-client/lib/models/experiment';
 import { OutputDescriptor } from 'tsp-typescript-client/lib/models/output-descriptor';
 import * as vscode from 'vscode';
 import { TraceViewerPanel } from '../../trace-viewer-panel/trace-viewer-webview-panel';
+import { TraceServerConnectionStatusService } from '../../utils/trace-server-status';
 import { getTraceServerUrl, getTspClientUrl } from '../../utils/tspClient';
 import { convertSignalExperiment } from 'vscode-trace-extension/src/common/signal-converter';
 
@@ -24,6 +25,7 @@ export class TraceExplorerAvailableViewsProvider implements vscode.WebviewViewPr
 
 	constructor(
 		private readonly _extensionUri: vscode.Uri,
+		private readonly _statusService: TraceServerConnectionStatusService,
 	) { }
 
 	public resolveWebviewView(
@@ -47,6 +49,12 @@ export class TraceExplorerAvailableViewsProvider implements vscode.WebviewViewPr
 	    // Handle messages from the webview
 	    webviewView.webview.onDidReceiveMessage(message => {
 	        switch (message.command) {
+	        case 'connectionStatus':
+	            if (message.data && message.data.status) {
+	                const status: boolean = JSON.parse(message.data.status);
+	                this._statusService.render(status);
+	            }
+	            return;
 	        case 'webviewReady':
 	            // Post the tspTypescriptClient
 	            webviewView.webview.postMessage({command: 'set-tspClient', data: getTspClientUrl()});
--- a/vscode-trace-extension/src/trace-explorer/opened-traces/trace-explorer-opened-traces-webview-provider.ts
+++ b/vscode-trace-extension/src/trace-explorer/opened-traces/trace-explorer-opened-traces-webview-provider.ts
@@ -5,6 +5,7 @@ import { Experiment } from 'tsp-typescript-client/lib/models/experiment';
 import * as vscode from 'vscode';
 import { convertSignalExperiment } from 'vscode-trace-extension/src/common/signal-converter';
 import { TraceViewerPanel } from '../../trace-viewer-panel/trace-viewer-webview-panel';
+import { TraceServerConnectionStatusService } from '../../utils/trace-server-status';
 import { getTraceServerUrl, getTspClientUrl } from '../../utils/tspClient';
 
 const JSONBig = JSONBigConfig({
@@ -21,6 +22,7 @@ export class TraceExplorerOpenedTracesViewProvider implements vscode.WebviewView
 
    	constructor(
 		private readonly _extensionUri: vscode.Uri,
+		private readonly _statusService: TraceServerConnectionStatusService,
 	) {}
 
 	private _onOpenedTracesWidgetActivated = (experiment: Experiment): void => this.doHandleTracesWidgetActivatedSignal(experiment);
@@ -76,6 +78,12 @@ export class TraceExplorerOpenedTracesViewProvider implements vscode.WebviewView
 	    // Handle messages from the webview
 	    webviewView.webview.onDidReceiveMessage(message => {
 	        switch (message.command) {
+	        case 'connectionStatus':
+	            if (message.data && message.data.status) {
+	                const status: boolean = JSON.parse(message.data.status);
+	                this._statusService.render(status);
+	            }
+	            return;
 	        case 'webviewReady':
 	            // Post the tspTypescriptClient
 	            webviewView.webview.postMessage({command: 'set-tspClient', data: getTspClientUrl()});
@@ -89,7 +97,7 @@ export class TraceExplorerOpenedTracesViewProvider implements vscode.WebviewView
 	        case 'reopenTrace':
 	            if (message.data && message.data.wrapper) {
 	                const experiment = convertSignalExperiment(JSONBig.parse(message.data.wrapper));
-	                const panel = TraceViewerPanel.createOrShow(this._extensionUri, experiment.name);
+	                const panel = TraceViewerPanel.createOrShow(this._extensionUri, experiment.name, this._statusService);
 	                panel.setExperiment(experiment);
 	                signalManager().fireExperimentSelectedSignal(experiment);
 	            }
@@ -113,6 +121,7 @@ export class TraceExplorerOpenedTracesViewProvider implements vscode.WebviewView
 	        }
 	        }
 	    }, undefined, this._disposables);
+
 	    signalManager().on(Signals.TRACEVIEWERTAB_ACTIVATED, this._onOpenedTracesWidgetActivated);
 	    signalManager().on(Signals.OPENED_TRACES_UPDATED, this._onOpenedTracesChanged);
 	    signalManager().on(Signals.EXPERIMENT_SELECTED, this._onExperimentSelected);
--- a/vscode-trace-extension/src/trace-viewer-panel/trace-viewer-webview-panel.ts
+++ b/vscode-trace-extension/src/trace-viewer-panel/trace-viewer-webview-panel.ts
@@ -11,7 +11,7 @@ const JSONBig = JSONBigConfig({
     useNativeBigInt: true,
 });
 
-// TODO: manage mutiple panels (currently just a hack around, need to be fixed)
+// TODO: manage multiple panels (currently just a hack around, need to be fixed)
 
 /**
  * Manages react webview panels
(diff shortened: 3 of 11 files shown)


bhufmann:
I going to review this today.

bhufmann (changes requested):
I noticed that when TSP queries in the trace panel fail due to non-connected server, the status in s not updated. 

With this current approach by polling in the provider only certain events will trigger that code path and the server status is not updated.

I think you should be able to use the ```RestClient.addConnectionStatusListener```. Each webview has an instance of the tspclient and also access the static connection listener. When sending a tsp message then it the listener implementation can notify via a message it's vscode extension counterpart and update the status line. I think if you add it to the common class `TspClientProvider` then you can share the code. Note that the `class` TraceViewerContainer doesn't use the `TspClientProvider` class, but you can change that at the same time to be able to use the connection listener implementation.

vscode-trace-extension/src/utils/trace-server-status.ts
```diff
@@ -0,0 +1,29 @@
...
+    public async renderStatus(): Promise<void> {
+        const response = await getTspClient().checkHealth();
+        const health = response.getModel();
+        let isUp = false;
+        if (health) {
+            isUp = health.status === 'UP';
+        }
+        if (isUp) {
+            this.statusBarItem.backgroundColor = new ThemeColor('statusBarItem.warningBackground');
+            this.statusBarItem.text = 'Trace Server: Found';
```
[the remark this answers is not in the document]
marco-miller: Good idea; thanks. I'm about to push an improved text as above. As for the tooltip, I'd prefer sticking to a found/not-found terminology, which I'm about to push as well. This is because the server might be online, albeit not found given some misconfiguration on extension's behalf (eventually).

vscode-trace-extension/src/utils/trace-server-status.ts
```diff
@@ -0,0 +1,29 @@
...
+        } else {
+            this.statusBarItem.backgroundColor = new ThemeColor('statusBarItem.errorBackground');
+            this.statusBarItem.text = 'Trace Server: Not Found';
```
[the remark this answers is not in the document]
marco-miller: Per my other related reply above; thanks again.

marco-miller:
This PR is ready for review again; thanks.

vscode-trace-webviews/src/common/vscode-message-manager.ts
```diff
@@ -40,9 +39,15 @@ export class VsCodeMessageManager extends Messages.MessageManager {
         vscode.postMessage({command: 'webviewReady'});
     }
 
+    notifyConnection(serverStatus: boolean): void {
+        const status: string = JSON.stringify(serverStatus);
```
bhufmann, line 43: You don't need to do JSON.stringify() and can remove it. The vscode postMessage() will do JSON.stringify() when sending and JSON.parse() when receiving the message. 

The reason why below JSONBig.stringify() is called is that the data structure that is sent contains BigInt and JSON.stringify() of postMessage() will cause and exception because the standard JSON.stringify() can't handle such special types (BigInt). So, in our code we do  JSONBig.stringify() to convert it to string, which then in postMessage() JSON.stringify() is called and on the receiving side, after JSON.parse() the trace extension has special code to recreate the data structure with BigInts

vscode-trace-extension/src/utils/trace-server-status.ts
```diff
@@ -0,0 +1,24 @@
...
+    public constructor(statusBarItem: StatusBarItem) {
+        this.statusBarItem = statusBarItem;
+        this.statusBarItem.hide();
+    }
+
+    public async render(status: boolean): Promise<void> {
+        if (status) {
+            this.statusBarItem.backgroundColor = new ThemeColor('statusBarItem.warningBackground');
+            this.statusBarItem.text = '$(check) Trace Server';
+            this.statusBarItem.tooltip = 'Trace Viewer: server found';
```
bhufmann, line 16: Should `server` be `Server`?

vscode-trace-extension/src/utils/trace-server-status.ts
```diff
@@ -0,0 +1,24 @@
...
+        } else {
+            this.statusBarItem.backgroundColor = new ThemeColor('statusBarItem.errorBackground');
+            this.statusBarItem.text = '$(error) Trace Server';
+            this.statusBarItem.tooltip = 'Trace Viewer: server not found';
```
bhufmann, line 20: Should `server` be `Server`?

vscode-trace-extension/src/trace-explorer/available-views/trace-explorer-available-views-webview-provider.ts
```diff
@@ -47,6 +49,12 @@ export class TraceExplorerAvailableViewsProvider implements vscode.WebviewViewPr
 	    // Handle messages from the webview
 	    webviewView.webview.onDidReceiveMessage(message => {
 	        switch (message.command) {
+	        case 'connectionStatus':
+	            if (message.data && message.data.status) {
+	                const status: boolean = JSON.parse(message.data.status);
```
bhufmann, line 54: You don't need JSON.parse(), after removing JSON.stringify() at sending time.

vscode-trace-extension/src/trace-explorer/opened-traces/trace-explorer-opened-traces-webview-provider.ts
```diff
@@ -76,6 +78,12 @@ export class TraceExplorerOpenedTracesViewProvider implements vscode.WebviewView
 	    // Handle messages from the webview
 	    webviewView.webview.onDidReceiveMessage(message => {
 	        switch (message.command) {
+	        case 'connectionStatus':
+	            if (message.data && message.data.status) {
+	                const status: boolean = JSON.parse(message.data.status);
```
bhufmann, line 83: You don't need JSON.parse().

vscode-trace-extension/src/trace-viewer-panel/trace-viewer-webview-panel.ts
```diff
@@ -171,6 +176,12 @@ export class TraceViewerPanel {
 	            	TraceViewerPanel.saveTraceCsv(message.payload.data, ((this._experiment !== undefined) ? this._experiment.name : 'trace')+'.csv');
 	            }
 	            return;
+	        case 'connectionStatus':
+	            if (message.data && message.data.status && this._statusService) {
+	                const status: boolean = JSON.parse(message.data.status);
```
bhufmann, line 181: You don't need JSON.parse().

bhufmann (changes requested):
Thanks for the updates. A few more small things.

vscode-trace-webviews/src/common/vscode-message-manager.ts
marco-miller: > You don't need to do JSON.stringify() and can remove it. The vscode postMessage() will do JSON.stringify() when sending and JSON.parse() when receiving the message.

Not necessarily, according to my local testing of both approaches. Here, posting this boolean without explicitly stringifying it leads to no status received on the other end (hence no rendering). This is what I briefly mention (why this) in the commit message and PR description. This is one of the few unexpected behaviours I had to figure out recently indeed.

> The reason why below JSONBig.stringify() is called is that the data structure that is sent contains BigInt and JSON.stringify() of postMessage() will cause and exception because the standard JSON.stringify() can't handle such special types (BigInt). So, in our code we do JSONBig.stringify() to convert it to string, which then in postMessage() JSON.stringify() is called and on the receiving side, after JSON.parse() the trace extension has special code to recreate the data structure with BigInts

Yes, I understood that but thanks.

vscode-trace-extension/src/utils/trace-server-status.ts
marco-miller: I'd prefer to keep `server`, here in this specific case given the colon use.

vscode-trace-extension/src/utils/trace-server-status.ts
marco-miller: Per above; thanks.

vscode-trace-extension/src/trace-explorer/available-views/trace-explorer-available-views-webview-provider.ts
marco-miller: Not necessarily, according to my local testing of both approaches. Here, not explicitly parsing led to the wrong status being rendered (misinterpreted in here). It ends up looking like we do have to explicitly stringify-then-parse the boolean value, across; so stated in the commit message and PR description.

vscode-trace-extension/src/trace-explorer/opened-traces/trace-explorer-opened-traces-webview-provider.ts
marco-miller: Per above; thanks.

vscode-trace-extension/src/trace-viewer-panel/trace-viewer-webview-panel.ts
marco-miller: Per above here too.

vscode-trace-webviews/src/common/vscode-message-manager.ts
bhufmann: You can remove it, it's not needed. Here do: 
```vscode.postMessage({command: 'connectionStatus', data: { serverStatus }});```

The receiving side needs to be updated as well. I'll put comment at one place on how.

File: vscode-trace-extension/src/trace-explorer/available-views/trace-explorer-available-views-webview-provider.ts (hunk 6 above)
Comment (bhufmann) [reply vscode-trace-extension/src/trace-explorer/available-views/trace-explorer-available-views-webview-provider.ts:54]: After removing the JSON conversion update as followed:
```
if (message.data && message.data.serverStatus !== undefined) {
    this._statusService.render(message.data.serverStatus);
}
```

bhufmann (changes requested):

vscode-trace-webviews/src/common/vscode-message-manager.ts
bhufmann: @marco-miller Your implementation works on my side too. So, since it seems to be a different behaviour without the extra marshalling, let's then merge it as is.

bhufmann (approved):
LGTM. Thanks for the contribution.

vscode-trace-extension/src/trace-explorer/available-views/trace-explorer-available-views-webview-provider.ts
bhufmann: Let's keep it as-is. It works fine.

PR #123, opened 2023-04-10
Configure integration tests

PatrickTasse:
Add mocha, @types/mocha and vscode-extenstion-tester to configuration.

Add ui-test script to build.

Remove clean from normal build script because .js files are needed when running ui-test.

Ignore and exclude test-resources folder from build and version control.


PatrickTasse:
Integration tests included with Playwright and openvscode-server instead in https://github.com/eclipse-cdt-cloud/vscode-trace-extension/pull/126.

PR #125, opened 2023-04-14
Added progress bar for file handler

ngondalia:
changes made:
- Implemented progress bar for when traces are opened.
- Added an output channel for the extension for logging purposes

This will provide the users with a progress bar when the trace file is opened. (Similar to how theia provides a progress bar)

Signed-off-by: Neel Gondalia ngondalia@blackberry.com

### DIFF
--- a/vscode-trace-extension/src/trace-explorer/trace-tree.ts
+++ b/vscode-trace-extension/src/trace-explorer/trace-tree.ts
@@ -7,14 +7,23 @@ import { ExperimentManager } from 'traceviewer-base/lib/experiment-manager';
 import { AnalysisProvider } from './analysis-tree';
 import { TraceViewerPanel } from '../trace-viewer-panel/trace-viewer-webview-panel';
 import { getTspClient } from '../utils/tspClient';
+import { traceLogger } from '../extension';
 
 const rootPath = path.resolve(__dirname, '../../..');
 
 const traceManager = new TraceManager(getTspClient());
 const experimentManager = new ExperimentManager(getTspClient(), traceManager);
 
+export enum ProgressMessages {
+    COMPLETE = 'Complete',
+    MERGING_TRACES = 'Merging trace(s)',
+    FINDING_TRACES = 'Finding trace(s)',
+    OPENING_TRACES = 'Opening trace(s)',
+    ROLLING_BACK_TRACES = 'Rolling back trace(s)'
+}
+
 export class TracesProvider implements vscode.TreeDataProvider<Trace> {
-    constructor(private workspaceRoot: string) { }
+    constructor(private workspaceRoot: string) {}
 
     getTreeItem(element: Trace): vscode.TreeItem {
         return element;
@@ -113,28 +122,40 @@ const openDialog = async (): Promise<vscode.Uri | undefined> => {
 };
 
 export const fileHandler = (analysisTree: AnalysisProvider) => async (context: vscode.ExtensionContext, traceUri: vscode.Uri | undefined): Promise<void> => {
+    // We need to resolve trace URI before starting the progress dialog because we rely on trace URI for dialog title
     if (!traceUri) {
         traceUri = await openDialog();
         if (!traceUri) {
-            console.log('Cannot open trace: invalid uri for trace', traceUri);
             return;
         }
     }
 
-    const uri: string = traceUri.path;
-    if (!uri) {
-        console.log('Cannot open trace: invalid uri for trace', traceUri);
-        return;
-    }
-    const name = uri.substring(uri.lastIndexOf('/') + 1);
-    const panel = TraceViewerPanel.createOrShow(context.extensionUri, name, undefined);
-    (async () => {
+    const resolvedTraceURI: vscode.Uri = traceUri;
+    vscode.window.withProgress({
+        location: vscode.ProgressLocation.Notification,
+        title: getProgressBarTitle(resolvedTraceURI),
+        cancellable: true
+    }, async (progress, token) => {
+        if (token.isCancellationRequested) {
+            progress.report({ message: ProgressMessages.COMPLETE, increment: 100 });
+            return;
+        }
 
+        const uri: string = resolvedTraceURI.path;
+        if (!uri) {
+            traceLogger.addLogMessage('Cannot open trace: could not retrieve path from URI for trace ' + resolvedTraceURI, fileHandler.name);
+            return;
+        }
+
+        const name = uri.substring(uri.lastIndexOf('/') + 1);
+        const panel = TraceViewerPanel.createOrShow(context.extensionUri, name, undefined);
+
+        progress.report({ message: ProgressMessages.FINDING_TRACES, increment: 10});
         /*
-     * TODO: use backend service to find traces
-     */
+        * TODO: use backend service to find traces
+        */
         const tracesArray: string[] = [];
-        const fileStat = await vscode.workspace.fs.stat(traceUri);
+        const fileStat = await vscode.workspace.fs.stat(resolvedTraceURI);
         if (fileStat) {
             if (fileStat.type === vscode.FileType.Directory) {
                 // Find recursively CTF traces
@@ -146,13 +167,38 @@ export const fileHandler = (analysisTree: AnalysisProvider) => async (context: v
             }
         }
 
+        if (tracesArray.length === 0) {
+            progress.report({ message: ProgressMessages.COMPLETE, increment: 100 });
+            traceLogger.addLogMessage('No valid traces found in the selected directory: ' + resolvedTraceURI, fileHandler.name);
+            panel.dispose();
+            return;
+        }
+
+        progress.report({ message: ProgressMessages.OPENING_TRACES, increment: 20 });
         const traces = new Array<TspTrace>();
         for (let i = 0; i < tracesArray.length; i++) {
             const traceName = tracesArray[i].substring(tracesArray[i].lastIndexOf('/') + 1);
             const trace = await traceManager.openTrace(tracesArray[i], traceName);
             if (trace) {
                 traces.push(trace);
+            } else {
+                traceLogger.addLogMessage('Failed to open trace: ' + traceName, fileHandler.name);
+                traceLogger.addLogMessage('There may be an issue with the server or the trace is invalid.', fileHandler.name);
             }
+        }
+
+        if (token.isCancellationRequested) {
+            rollbackTraces(traces, 20, progress);
+            progress.report({ message: ProgressMessages.COMPLETE, increment: 50 });
+            panel.dispose();
+            return;
+        }
+
+        progress.report({ message: ProgressMessages.MERGING_TRACES, increment: 40 });
+        if (traces === undefined || traces.length === 0) {
+            progress.report({ message: ProgressMessages.COMPLETE, increment: 30 });
+            panel.dispose();
+            return;
         }
 
         const experiment = await experimentManager.openExperiment(name, traces);
@@ -164,7 +210,27 @@ export const fileHandler = (analysisTree: AnalysisProvider) => async (context: v
             }
         }
 
-    })();
+        if (token.isCancellationRequested) {
+            if (experiment) {
+                experimentManager.deleteExperiment(experiment.UUID);
+            }
+            rollbackTraces(traces, 20, progress);
+            progress.report({ message: ProgressMessages.COMPLETE, increment: 10 });
+            panel.dispose();
+            return;
+        }
+        progress.report({ message: ProgressMessages.COMPLETE, increment: 30 });
+    });
+};
+
+const rollbackTraces = async (traces: Array<TspTrace>, progressIncrement: number, progress: vscode.Progress<{
+    message: string | undefined;
+    increment: number | undefined;
+}>) => {
+    progress.report({ message: ProgressMessages.ROLLING_BACK_TRACES, increment: progressIncrement});
+    for (let i = 0; i < traces.length; i++) {
+        await traceManager.deleteTrace(traces[i].UUID);
+    }
 };
 
 /*
@@ -204,3 +270,10 @@ const isCtf = async (directory: string): Promise<boolean> => {
     }
     return false;
 };
+
+function getProgressBarTitle(traceUri: vscode.Uri | undefined): string {
+    if (!traceUri || !traceUri.path) {
+        return 'undefined';
+    }
+    return traceUri.path.substring(traceUri.path.lastIndexOf('/') + 1);
+}


ngondalia:
Screenshot from 2023-04-14 15-35-11

vscode-trace-extension/src/utils/traceExtensionLogger.ts
```diff
@@ -0,0 +1,91 @@
+/***************************************************************************************
+ * Copyright (c) 2023 BlackBerry Limited and others.
+ *
+ * Licensed under the MIT license. See LICENSE file in the project root for details.
+ ***************************************************************************************/
+
+import * as vscode from 'vscode';
+
+export class traceExtensionLogger {
```
bhufmann, line 9: class name start with upper case: `TraceExtensionLogger"

vscode-trace-extension/src/utils/traceExtensionLogger.ts
```diff
@@ -0,0 +1,91 @@
+/***************************************************************************************
```
bhufmann, line 1: Please rename `traceExtensionLogger.ts` to `trace-extension-logger.ts` to be consistent with other source code files

Most of the methods are not used. Could you please add some description to the message to know what each method does?

Questions:
- Does vscode make sure that panel is not overflowing?
- Does the panel need to be hidden or disposed at some point?

bhufmann (changes requested):
Thanks for the contribution! Useful addition. I have some comments though.

vscode-trace-extension/src/utils/traceExtensionLogger.ts
ngondalia: I have added descriptions for the methods in the Logger.

Does vscode make sure that panel is not overflowing?
-> yes. There is also a method to clear the log channel both from the API and through the UI

 Does the panel need to be hidden or disposed at some point?
-> I have it disposing on extension deactivation now. Reason being my thinking is that this channel can be used by all the modules of the extension, not just the file handler. As for showing/hiding it, the current API only allows hiding the entire outputs panel, and not the individual channels. Hiding individual channels and replacing it with the previously shown channel would be more ideal but its the current limitation of the vscode api. For now, we can just show the output log channel whenever we log something to it and not worry about hiding it. Other extensions that use output channels would do the same.

->The alternatives are always keeping it hidden, and the user can decide when they want to see the trace extension channel by selecting it from the dropdown menu in the outputs panel. The other option is actually using the hide channel api call, which will hide the entire outputs panel (not my preference).

WDYT?

vscode-trace-extension/src/trace-explorer/trace-tree.ts
```diff
@@ -113,31 +114,47 @@ const openDialog = async (): Promise<vscode.Uri | undefined> => {
...
 export const fileHandler = (analysisTree: AnalysisProvider) => async (context: vscode.ExtensionContext, traceUri: vscode.Uri | undefined): Promise<void> => {
+    // Show the log output channel to the user while the trace is being processed
+    traceLogger.showLog();
+
+    // We need to resolve trace URI before starting the progress dialog because we rely on trace URI for dialog title
     if (!traceUri) {
         traceUri = await openDialog();
         if (!traceUri) {
-            console.log('Cannot open trace: invalid uri for trace', traceUri);
+            traceLogger.addLogMessage('Cannot open trace: invalid uri for trace ' + traceUri, fileHandler.name);
```
bhufmann, line 124: This is not an error and we should not log it. traceUri is undefined when the user cancels the dialog.

vscode-trace-extension/src/utils/traceExtensionLogger.ts
bhufmann: I think opening the log window everytime we log something might be a bit disruptive, because I would have to close it everytime manually when I want to get more vertical space. I understand that it would be less discoverable for end-users, but I think it's better. I don't think other extension open it automatically.

vscode-trace-extension/src/utils/trace-extension-logger.ts
```diff
@@ -0,0 +1,141 @@
...
+      return this.addLogMessage(fullMessage);
+  }
+
+  /**
+   * Appends the message with a timestamp to the log channel
+   * @param message message to be added to the log channel
+   */
+  private addLogPanelMessage(message: string) {
+      if (!this.channelDisposed) {
+          this.logChannel.append(`[${new Date().toLocaleTimeString('en-US', {hour12: false})}] ${message}\n`);
```
bhufmann, line 138: We should log also the time. I see other plug-ins doing the same. I will be useful if we log more things and want to trouble-shoot.

bhufmann (changes requested):

vscode-trace-extension/src/utils/traceExtensionLogger.ts
ngondalia: yeah i agree with that. We can just log to it, and if user wants to inspect it, they can open the channel.

vscode-trace-extension/src/utils/trace-extension-logger.ts
ngondalia: It does have time.. did you mean date?

ss1

bhufmann: Oh, true, the date is missing not the time. Right now we don't need more details. But you could change it right away for the future.  I checked a few outputs from other extension and most show something like this: 2023-04-20 13:37:24.500.

ngondalia: I have modified the timestamp to be in the same format as the one you mentioned

vscode-trace-extension/src/utils/trace-extension-logger.ts
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  /**
+   * Appends the message with a timestamp to the log channel
+   * @param message message to be added to the log channel
+   */
+  private addLogPanelMessage(message: string) {
+      if (!this.channelDisposed) {
+          const d = new Date();
+          d.getMilliseconds();
```
bhufmann, line 139: the last 2 lines are not needed

vscode-trace-extension/src/utils/trace-extension-logger.ts
bhufmann, line 1: The file uses other format than other source code file, e.g. 4 space for indention. Could you please format the file? Using the vscode / vscodium formatter should work.

bhufmann (changes requested):
Works fine Small nits left to fix.

bhufmann (approved):
Looks good to me. Thanks for the contribution.

vscode-trace-extension/src/extension.ts
```diff
@@ -6,8 +6,12 @@ import { TraceExplorerAvailableViewsProvider } from './trace-explorer/available-
 import { TraceExplorerOpenedTracesViewProvider } from './trace-explorer/opened-traces/trace-explorer-opened-traces-webview-provider';
 import { fileHandler, openOverviewHandler, resetZoomHandler } from './trace-explorer/trace-tree';
 import { updateTspClient } from './utils/tspClient';
+import { TraceExtensionLogger } from './utils/trace-extension-logger';
```
marco-miller, line 9: Can we split this PR into at least two commits? Say, one for this logging added, and another for the progress bar addition. If there are refactorings not pertaining to either one of these two main topics, then that should be at least yet another commit.

Doing so enables reviewability, and revertabiliy (should either one of the main changes be deemed problematic after merges). If they all sit in the same commit, then one has to revert everything in case of only some revertworthy bug(s).

Now, if a commit depends on (or builds on top of) another, then it can simply be based on the latter, as usual with git commits indeed.

vscode-trace-extension/src/trace-explorer/trace-tree.ts
```diff
@@ -164,7 +208,21 @@ export const fileHandler = (analysisTree: AnalysisProvider) => async (context: v
             }
         }
 
-    })();
+        if (token.isCancellationRequested) {
+            if (experiment) {
+                experimentManager.deleteExperiment(experiment.UUID);
+            }
+            // Rollback traces
+            progress.report({ message: 'Rolling back trace(s)',  increment: 20});
```
marco-miller, line 216: Could this duplicating message string be a reused constant instead? So that we consistently maintain it if need be; we could otherwise forget one. This may apply also to other non-trivial messages in this change.

In fact, maybe rather extract this whole duplicated rollbacking logic in a reused function?

vscode-trace-extension/src/trace-explorer/trace-tree.ts
```diff
@@ -146,15 +159,46 @@ export const fileHandler = (analysisTree: AnalysisProvider) => async (context: v
...
                 traces.push(trace);
+            } else {
+                traceLogger.addLogMessage('Failed to open trace: ' + traceName, fileHandler.name);
+                traceLogger.addLogMessage('There may be an issue with the server or the trace is invalid.', fileHandler.name);
+                invalidTraces.push(traceName);
             }
         }
 
+        if (token.isCancellationRequested) {
+            // Rollback traces
```
marco-miller, line 185: I'd remove this one-liner comment, as we already state the matter thanks to line 186's message right below.

vscode-trace-extension/src/trace-explorer/trace-tree.ts
```diff
@@ -164,7 +208,21 @@ export const fileHandler = (analysisTree: AnalysisProvider) => async (context: v
             }
         }
 
-    })();
+        if (token.isCancellationRequested) {
+            if (experiment) {
+                experimentManager.deleteExperiment(experiment.UUID);
+            }
+            // Rollback traces
```
marco-miller, line 215: Same as above, here too.

vscode-trace-extension/src/utils/trace-extension-logger.ts
```diff
@@ -0,0 +1,182 @@
...
+     */
+    logError(e: Error): void {
+        this.addLogMessage(e.message);
+        if (e.stack) {
+            this.addLogMessage(e.stack);
+        }
+    }
+
+    /**
+     * Shows a information message to user as a notification and
```
marco-miller, line 91: "Shows `an` information message", rather.

vscode-trace-extension/src/utils/trace-extension-logger.ts
```diff
@@ -0,0 +1,182 @@
...
+     * adds the message to log the log channel
```
marco-miller, line 92: "message to the log channel", rather?

vscode-trace-extension/src/utils/trace-extension-logger.ts
```diff
@@ -0,0 +1,182 @@
...
+     */
+    async showInfo(message: string, options?: vscode.MessageOptions): Promise<void> {
+        const fullMessage = 'Info: ' + message;
+        vscode.window.showInformationMessage(fullMessage, options || { modal: false });
+        return this.addLogMessage(fullMessage);
+    }
+
+    /**
+     * Shows a warning message to user as a notification and
+     * adds the message to log the log channel
```
marco-miller, line 106: Same as earlier.

vscode-trace-extension/src/utils/trace-extension-logger.ts
```diff
@@ -0,0 +1,182 @@
...
+     */
+    async showWarning(message: string, options?: vscode.MessageOptions): Promise<void> {
+        const fullMessage = 'Warning: ' + message;
+        vscode.window.showWarningMessage(fullMessage, options || { modal: false });
+        return this.addLogMessage(fullMessage);
+    }
+
+    /**
+     * Shows an error message to user as a notification and
+     * adds the message to log the log channel
```
marco-miller, line 120: Same here too.

vscode-trace-extension/src/utils/trace-extension-logger.ts
```diff
@@ -0,0 +1,182 @@
...
+     */
+    async showError(message: string, options?: vscode.MessageOptions): Promise<void> {
+        const fullMessage = 'Error: ' + message;
+        vscode.window.showErrorMessage(fullMessage, options || { modal: false });
+        return this.addLogMessage(fullMessage);
+    }
+
+    /**
+     * Appends the message with a timestamp to the log channel
+     * @param message message to be added to the log channel
```
marco-miller, line 134: The previous function docs have a blank line before their `@param` tags.

Applies to the following ones below and down.

marco-miller (changes requested):
Thanks, yes; some more comments please.

vscode-trace-extension/src/extension.ts
ngondalia: Done! https://github.com/eclipse-cdt-cloud/vscode-trace-extension/pull/129

Once the logging commit is merged, i will update this PR. Thanks!

ngondalia:
Depends on #129

marco-miller:
> Depends on #129

You could also base this commit on #129's, say, in that PR (branch) solely. Thus having two commits in that PR. Then amend the PR description so it combines both speeches, logically speaking. But each commit would remain separate with its focused commit message, indeed. Us reviewers may then review /try each commit, step-wise. This approach has the advantage of preserving the dependency context altogether, while allowing for single commits to exist and be reviewable.

You'd also eventually need to individually Resolve or reply to each open comment above; thanks. This is so that I (the Reviewer here) know when to review again.

ngondalia:
@bhufmann @marco-miller It should be good now. Please review whenever you get a chance! Thanks!

vscode-trace-extension/src/trace-explorer/trace-tree.ts
```diff
@@ -164,7 +210,27 @@ export const fileHandler = (analysisTree: AnalysisProvider) => async (context: v
...
+            rollbackTraces(traces, 20, progress);
+            progress.report({ message: ProgressMessages.COMPLETE, increment: 10 });
+            panel.dispose();
+            return;
+        }
+        progress.report({ message: ProgressMessages.COMPLETE, increment: 30 });
+    });
+};
+
+const rollbackTraces =async (traces: Array<TspTrace>, progressIncrement: number, progress: vscode.Progress<{
```
marco-miller, line 226: `= async`

vscode-trace-extension/src/trace-explorer/trace-tree.ts
```diff
@@ -164,7 +210,27 @@ export const fileHandler = (analysisTree: AnalysisProvider) => async (context: v
...
+    message?: string | undefined;
```
marco-miller, line 227: `message:` would suffice, as `undefined` is already specified (both being mutually redundant here).

vscode-trace-extension/src/trace-explorer/trace-tree.ts
```diff
@@ -164,7 +210,27 @@ export const fileHandler = (analysisTree: AnalysisProvider) => async (context: v
...
+    increment?: number | undefined;
```
marco-miller, line 228: `increment:`, per above.

vscode-trace-extension/src/trace-explorer/trace-tree.ts
```diff
@@ -164,7 +210,27 @@ export const fileHandler = (analysisTree: AnalysisProvider) => async (context: v
...
+}>) => {
+    progress.report({ message: ProgressMessages.ROLLING_BACK_TRACES,  increment: progressIncrement});
```
marco-miller, line 230: `, increment` (space doubled).

marco-miller (changes requested):
A few minor comments; thanks.

marco-miller (approved):
LGTM now, thanks. Letting @bhufmann do the second review.

bhufmann (approved):
Looks good to me! Thanks for the contribution.

Issue #71, opened 2022-10-11
Fix style of DATA_TREE component


bhufmann:
The style of the DATA_TREE component seems to be off and not the same as in `theia-trace-extension`. The styles need to fixed. 

`vscode-trace-extension`
![image](https://user-images.githubusercontent.com/413424/195191380-24cfccbc-6b6b-49fd-ac03-cfc712f51a7b.png)

In comparison the `theia-trace-extension`
![image](https://user-images.githubusercontent.com/413424/195191556-b57cefdb-52f8-40a9-bbe5-8d3b9f323b4f.png)

bhufmann:
It turns out any tree implementation (e.g. tree in xy-output-component) has the same issue. Fixing it for one component will fix it for the other one as well because they share code.

PR #136, opened 2023-05-03
Integrate codicons with webviews

### LINKED ISSUE #89: Integrate codicons with webviews
Codicons are not integrated in webviews by default, due to certain icons might be missing. The following issue talks about how to let the webview load VS Code's bundled codicons: https://github.com/microsoft/vscode/issues/95199.

hoangphamEclipse:
This commit adds a reference to the codicons CSS file to enable codicons inside the TraceViewPanel of the vs-code trace extension. It follows the example provided in the [vs-code webview sample]( https://github.com/microsoft/vscode-extension-samples/tree/main/ webview-codicons-sample).

Fixes #89.


bhufmann (changes requested):
The code seems right. However, it doesn't fix the missing icon described in #71, i.e. the filter icon is still now show. 

Also, should we do the same for the other webviews (e.g. opened traces) in this project?

hoangphamEclipse:
Hi Bernd, I tested the code on my end, I have the filter icons displayed on the DATA_TREE view mentioned in #71. 

image

bhufmann (changes requested):
It works now when running it from the development environment using the launch configuration. However, after building the vsix (yarn vsce:package) and installing it in VsCode using install from vsix, then it doesn't work, i.e. the filter icon is not shown. I suspect that the packaging has to be adapted as well.

hoangphamEclipse:
Looks like it is a known-issue with `Codicons`. You can read about it [here](https://github.com/microsoft/vscode-extension-samples/issues/692). There is a work-around. I will try to see if that works.

hoangphamEclipse:
Turns out the [support for yarn workspace is not quite there yet](https://github.com/microsoft/vscode-vsce/issues/300) for `vsce package`. Will continue to look for work around.

hoangphamEclipse:
I used the [`copy-webpack-plugin`](https://webpack.js.org/plugins/copy-webpack-plugin/) to copy the Codicons CSS and font files automatically during the packaging process. The files are stored within `lib/codicons`. Please let me know we need to change file location.

bhufmann (approved):
It looks good to me. I tested with VsCode and Theia and the icons are now visible.

Issue #15, opened 2021-02-16
Migrate start and stop of trace server using the backend


bhufmann:
In the `theia-trace-extension` it's possible to start/stop the trace server from the Theia backend. The command `Start trace server` and respectively `Stop trace server` exist that starts/stops the trace server configured by the preferences in the `theia-trace-extension`. This functionality was discussed in https://github.com/theia-ide/theia-trace-extension/issues/53 which references the initial implementation of the feature. The start command is also triggered when trying to open a trace with the `theia-trace-extension`.

For the `vscode-trace-extension`, a similar support is needed to start/stop the server for both the local and remote cases.

Note: Installing the `vscode-trace-extension` in VsCode and Theia applications should work by the update.

Also supporting a `TRACE_SERVER_URL` [environment variable](https://github.com/eclipse-cdt-cloud/theia-trace-extension#change-the-trace-server-url) is out of this scope for now, if ever required (not yet).

PR #142, opened 2023-05-16
extension: Start trace server before opening trace

marco-miller:
Start the trace server if known as stopped, and if [1] is installed, before opening traces.

[1] https://github.com/eclipse-cdt-cloud/vscode-trace-server

Contributes to fixing Issue #15.

### DIFF
--- a/vscode-trace-extension/src/extension.ts
+++ b/vscode-trace-extension/src/extension.ts
@@ -41,6 +41,7 @@ export function activate(context: vscode.ExtensionContext): void {
     // TODO: For now, a different command opens traces from file explorer. Remove when we have a proper trace finder
     const fileOpenHandler = fileHandler(analysisProvider);
     context.subscriptions.push(vscode.commands.registerCommand('traces.openTraceFile', file => {
+        startTraceServerIfAvailable();
         fileOpenHandler(context, file);
     }));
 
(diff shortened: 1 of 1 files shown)


marco-miller:
Comment from internal, from @bhufmann:
- "let's backlog and discuss the way forward.
- Some downstream projects of the `theia-trace-extension` introduce a command to open the traces without starting the trace server automatically,
- because they have [their] own way to start/stop the trace server
- and want to by-pass the automatic start and stop."

bhufmann (commented):
I think this is also a question to the community if that is a general behaviour that is expected. In our use case it's a yes, but I know from the theia-trace-extension adopters, that they have other means to start the server for the end-user and there is an additional open command that doesn't try to start the server.

bhufmann (commented):
With the current implementation, it requires the vscode-traceserver to be installed to have the commands available. This adds a tight coupling to vscode-traceserver. What about an api in vscode-trace-extension that other extensions can register a start and stop callback. This will allow other implementations that vscode-traceserver. Just a thought.

AlexDoanBB:
hi, it seems that this feature has been backlogged. Is there any plan to add an api for other extensions to register start and stop callback? I'd like to build upon https://github.com/eclipse-cdt-cloud/vscode-trace-extension/pull/149 for this feature. Please let me know your thoughts, thanks

bhufmann:
This one can be closed due to #155. Thanks for the proposal.

PR #144, opened 2023-05-23
extension: Support hidden server status commands

marco-miller:
Add two commands for a sibling extension such as [1] to use.
- These two commands are hidden from the Command Palette,
- as meant for [1] to update the Trace Server status bar item upon confirmed server start/stop.

[1] https://github.com/eclipse-cdt-cloud/vscode-trace-server

Contributes to fixing Issue #15.

### DIFF
--- a/vscode-trace-extension/package.json
+++ b/vscode-trace-extension/package.json
@@ -80,6 +80,14 @@
         "command": "traceViewer.shortcuts",
         "title": "Trace Viewer Keyboard and Mouse Shortcuts",
         "icon": "$(info)"
+      },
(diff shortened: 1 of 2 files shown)


bhufmann (approved):
It works with the code in the `vscode-trace-server` start and stop feature and the status bar is updated accordingly. Interesting approach to have a hidden command. 

Alternatively, there could be an explicit API to call from outside. But, I don't see any issues with your approach and we can keep it.

bhufmann (approved):
LGTM. Thanks!

PR #145, opened 2023-05-26
Update and remove unused CSS variables to match new style definition

hoangphamEclipse:
This commit updates the CSS variables of the vscode-trace-extension to match the new CSS definition of the traceviewer-base package. This commit also removes CSS variables that are not used in both of the vscode and theia trace extensions.


bhufmann (approved):
LGTM. Thanks!

PR #149, opened 2023-06-06
Implemented external API

### LINKED ISSUE #140: Provide API to allow the capability to extend the trace extension
Currently, there is no way for another extension to customize the Trace Viewer and add features onto the existing views. Adding this capability would be very beneficial for cases where someone wants to add another view to the trace viewer that is specific to their use case and would not be acceptable to add it to the base extension. For example, some trace types may want to show raw bytes of an event; They can contribute a Raw Bytes view to the trace viewer from their extension, and using the API provided by vscode-trace-extension, it can communicate the selected experiment, selected event, etc. The calling extension can then update their view accordingly.

Adding an API for an extension is quite simple with the VSCode API:
https://code.visualstudio.com/api/references/vscode-api#extensions

The alternative is to use VSCode commands, but it may not be scalable as the API grows larger.

For the initial API, I am interested in obtaining the experiment UUID of the current panel open in Trace Viewer. This issue can track the discussion of this feature, and the work for providing an API.

ngondalia:
This commit provides an external API that can be used by other extensions to communicate with the base trace extension. The current API provides a way to retrieve the active experiment as well as fire events and register callbacks. This is the initial API and will be further expanded in future commits.

Fixes #140

### DIFF
--- a/vscode-trace-extension/src/external-api/external-api.ts
+++ b/vscode-trace-extension/src/external-api/external-api.ts
@@ -1,0 +1,64 @@
+/***************************************************************************************
+ * Copyright (c) 2023 BlackBerry Limited and others.
+ *
+ * Licensed under the MIT license. See LICENSE file in the project root for details.
+ ***************************************************************************************/
+import { Experiment } from 'tsp-typescript-client/lib/models/experiment';
+import { TraceViewerPanel } from '../trace-viewer-panel/trace-viewer-webview-panel';
+import * as vscode from 'vscode';
+import { traceExtensionWebviewManager } from '../extension';
+
+export interface ExternalAPI {
+    getActiveExperiment(): Experiment | undefined,
+    getActiveWebviewPanels(): { [key: string]: TraceViewerPanel | undefined; },
+    getActiveWebviews(): vscode.WebviewView[],
+    onWebviewCreated(listener: (data: vscode.WebviewView) => void): void,
+    onWebviewPanelCreated(listener: (data: vscode.WebviewPanel) => void): void,
+}
+
+export const traceExtensionAPI: ExternalAPI = {
+    /**
+     * Retrieves the currently active experiment
+     *
+     * @returns Experiment if one is currently active, otherwise undefined
+     */
+    getActiveExperiment(): Experiment | undefined {
+        return TraceViewerPanel.getCurrentExperiment();
+    },
+
+    /**
+     * Retrieves active trace panels
+     *
+     * @returns Key value pairs where the value is of TraceViewerPanel type if panel with that key is active, otherwise undefined
+     */
+    getActiveWebviewPanels(): { [key: string]: TraceViewerPanel | undefined; } {
+        return TraceViewerPanel.activePanels;
+    },
+
+    /**
+     * Retrieves active webviews
+     *
+     * @returns List of active webviews
+     */
+    getActiveWebviews(): vscode.WebviewView[] {
+        return traceExtensionWebviewManager.getAllActiveWebviews();
+    },
+
+    /**
+     * Registers an event listener for onWebviewCreated event
+     *
+     * @param listener event listener
+     */
+    onWebviewCreated(listener: (data: vscode.WebviewView) => void): void {
+        traceExtensionWebviewManager.onWebviewCreated(listener);
+    },
+
+    /**
+     * Registers an event listener for onWebviewPanelCreated event
+     *
+     * @param listener event listener
+     */
+    onWebviewPanelCreated(listener: (data: vscode.WebviewPanel) => void): void {
+        traceExtensionWebviewManager.onWebviewPanelCreated(listener);
+    }
+};
--- a/vscode-trace-extension/src/trace-explorer/opened-traces/trace-explorer-opened-traces-webview-provider.ts
+++ b/vscode-trace-extension/src/trace-explorer/opened-traces/trace-explorer-opened-traces-webview-provider.ts
@@ -7,6 +7,7 @@ import { TraceViewerPanel } from '../../trace-viewer-panel/trace-viewer-webview-
 import { TraceServerConnectionStatusService } from '../../utils/trace-server-status';
 import { getTraceServerUrl, getTspClientUrl } from '../../utils/tspClient';
 import { VSCODE_MESSAGES } from 'vscode-trace-common/lib/messages/vscode-message-manager';
+import { traceExtensionWebviewManager } from 'vscode-trace-extension/src/extension';
 
 const JSONBig = JSONBigConfig({
     useNativeBigInt: true,
@@ -20,7 +21,7 @@ export class TraceExplorerOpenedTracesViewProvider implements vscode.WebviewView
 	private _disposables: vscode.Disposable[] = [];
 	private _selectedExperiment: Experiment | undefined;
 
-   	constructor(
+	constructor(
 		private readonly _extensionUri: vscode.Uri,
 		private readonly _statusService: TraceServerConnectionStatusService,
 	) {}
@@ -69,6 +70,7 @@ export class TraceExplorerOpenedTracesViewProvider implements vscode.WebviewView
 	    };
 
 	    webviewView.webview.html = this._getHtmlForWebview(webviewView.webview);
+	    traceExtensionWebviewManager.fireWebviewCreated(webviewView);
 
 	    // Handle messages from the webview
 	    webviewView.webview.onDidReceiveMessage(message => {
--- a/vscode-trace-extension/src/utils/trace-extension-webview-manager.ts
+++ b/vscode-trace-extension/src/utils/trace-extension-webview-manager.ts
@@ -1,0 +1,78 @@
+/***************************************************************************************
+ * Copyright (c) 2023 BlackBerry Limited and others.
+ *
+ * Licensed under the MIT license. See LICENSE file in the project root for details.
+ ***************************************************************************************/
+import * as vscode from 'vscode';
+
+/**
+ * Manages webview panels and webviews creation events
+ */
+export class TraceExtensionWebviewManager {
+    private webviews: vscode.WebviewView[] = [];
+    private webviewCreated: vscode.EventEmitter<vscode.WebviewView> = new vscode.EventEmitter();
+    private webviewPanelCreated: vscode.EventEmitter<vscode.WebviewPanel> = new vscode.EventEmitter();
+    private isManagerDisposed = false;
+
+    getAllActiveWebviews(): vscode.WebviewView[] {
+        if (!this.isDisposed()) {
+            return this.webviews;
+        }
+        return [];
+    }
+
+    fireWebviewCreated(_webview: vscode.WebviewView): void {
+        if (!this.isDisposed()) {
+            this.addWebview(_webview);
+            this.webviewCreated.fire(_webview);
+        }
+    }
+
+    fireWebviewPanelCreated(_webviewPanel: vscode.WebviewPanel): void {
+        if (!this.isDisposed()) {
+            this.webviewPanelCreated.fire(_webviewPanel);
+        }
+    }
+
+    onWebviewCreated(listener: (data: vscode.WebviewView) => unknown): void {
+        if (!this.isDisposed()) {
+            this.webviewCreated.event(listener);
+        }
+    }
+
+    onWebviewPanelCreated(listener: (data: vscode.WebviewPanel) => unknown): void  {
+        if (!this.isDisposed()) {
+            this.webviewPanelCreated.event(listener);
+        }
+    }
+
+    dispose(): void {
+        if (!this.isDisposed()) {
+            this.webviews = [];
+            this.webviewCreated.dispose();
+            this.webviewPanelCreated.dispose();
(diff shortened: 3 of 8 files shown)


bhufmann (commented):
Thanks for the contribution. I was able to use the external API in respect with getting the active trace.

I'm not clear on how to use the signal interface. @ngondalia, could you please walk me through what needs to implemented from an another extension using a simple example?

ngondalia:
const ext = vscode.extensions.getExtension("tracecompass-community.vscode-trace-extension");
const importedApi = ext.exports;

// The following gets the keys of event emitters that are registered to TraceExtensionEventManager
importedApi.getEventEmitterIds();

/*
 * The only registered event emitter right now is the "traceExplorer.openedTracesView.register_callback" - which onEvent 
 *  registers the callback for a given signal in the OpenedTracesView
 */
importedApi.fireEventEmitter("traceExplorer.openedTracesView.register_callback", "experimentSelected", (data: any) => {
            vscode.window.showInformationMessage("Event Fired: exp selected");
});

//Now the provided callback every time the "experimentSelected" is propagated in the Opened traces view

vscode-trace-extension/src/external-api/external-api.ts
```diff
@@ -0,0 +1,43 @@
...
+        return TraceViewerPanel.getCurrentExperiment();
+    },
+
+    /**
+     * Fires the event emitter with the provided ID
+     *
+     * @param eventId ID of the event emitter to fire
+     * @param data event object
+     */
+    fireEventEmitter(eventId: string, data: unknown): void {
```
bhufmann, line 32: It took me a while to understand how it's implemented. When an extension wants to be notified about a signal received by a webview, it uses this API to fire an event through the eventEmitter that will add a callback to a webview which is called when a signal is received by the webview and contains the signal name provided (when calling fireEventEmitter). I guess EventEmitter is the standard way to communicate between webview and that's why you used this approach. Or was that another reason. Please let me know. 

Second questions, should `data` be of type `SignalCallbackInterface`? Or is it set to `unknown` so that we can use if for other use cases than signal callbacks? Do you have one in mind?

bhufmann (commented):
I was able to test the signal call back interface. I had to modify your example a bit to make it work:
```
importedApi.fireEventEmitter("traceExplorer.openedTracesView.register_callback", 
			{
				signal: "experimentSelected", 
				callback: (data: any) => {
            		vscode.window.showInformationMessage("Event Fired: exp selected");
				}
			}
		);
```
I tried to add a similar callback mechanism to available views webview provider (e.g. for `output-added` signal). 

I noticed when the webviews are minimised the callback is not called anymore because the webviews are disposed. The callbacks are only called when the webview is visible. For the `experimentSelected` signal you can reproduce it when have at least 2 traces open, minimise the opened traces view and then switch traces panels. Not sure if this is an issue.

When the callback is called form when the message is received, it will forward the data structure. For `experimentSelected` it will contain the data about the Experiment. The actual `Experiment` class couldn't be used in the signal directly, because the start/end time are of type `bitint` which cannot be serialized to JSON by default. So, what the `vscode-trace-extension` does is to use JSONbigInt to serialize it to a string before sending and on the receiving side deserializes the string to the `Experiment` type re-creating start/end time as `bigint`. No, with this callback mechnism the receiving side needs to do the same thing. Right now there is no generic utility to be re-used. Maybe https://github.com/eclipse-cdt-cloud/tsp-typescript-client/pull/56 could help, but the project hasn't decided to go with it.

Another comment is, that with this callback feature, the messages and their data structure become also API so that adopter extension can rely on. 

We should add some documentation on the external API and how to use it. Your example would be good candidate to show how to use it. 

I wrote quite a bit text as reply. Could you please look at the comments and let me know what you think and how we should proceed. 

Thanks
Bernd

ngondalia:
> I noticed when the webviews are minimised the callback is not called anymore because the webviews are disposed. The callbacks are only called when the webview is visible. For the `experimentSelected` signal you can reproduce it when have at least 2 traces open, minimise the opened traces view and then switch traces panels. Not sure if this is an issue.

Yeah this is expected because the only place from which the "register_callback" event is registered with the traceExtensionEventManager, is TraceExplorerOpenedTracesViewProvider. But technically, you can do the same from all the webviews or for the extension itself (register a "register_callback" event with the traceExtensionManager).

ngondalia:
> When the callback is called form when the message is received, it will forward the data structure. For `experimentSelected` it will contain the data about the Experiment. The actual `Experiment` class couldn't be used in the signal directly, because the start/end time are of type `bitint` which cannot be serialized to JSON by default. So, what the `vscode-trace-extension` does is to use JSONbigInt to serialize it to a string before sending and on the receiving side deserializes the string to the `Experiment` type re-creating start/end time as `bigint`. No, with this callback mechnism the receiving side needs to do the same thing. Right now there is no generic utility to be re-used. Maybe [eclipse-cdt-cloud/tsp-typescript-client#56](https://github.com/eclipse-cdt-cloud/tsp-typescript-client/pull/56) could help, but the project hasn't decided to go with it.

I think having a generic utility definitely should be the way going forward. Currently however, the calling extension has to handle this similarly to how the vscode-trace-extension does it.

ngondalia:
> Another comment is, that with this callback feature, the messages and their data structure become also API so that adopter extension can rely on.
> 
> We should add some documentation on the external API and how to use it. Your example would be good candidate to show how to use it.

Yeah I agree. I will add more helpers in the API to create these data structures. I will also add a README in the following commit. Thanks!

vscode-trace-extension/src/utils/trace-extension-event-manager.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * Attaches a listener for an event based on the event emitter id
+     *
+     * @param id event emitter Id
+     * @param callback function to call when the event is triggered
+     */
+    onEvent(id: string, callback: (data: unknown) => void): void {
+        this.getEventEmitter(id)?.event(callback);
+    }
```
colin-grant-work, line 68: Is there a reason not to return the `Disposable` that is returned by the underlying `event` call?

And if the emitter doesn't currently exist, should it be created? Or should the caller at least be informed in some way that their subscription failed?

vscode-trace-extension/src/utils/trace-extension-event-manager.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * Register an event emitter with the provided Id
+     *
+     * @param id event emitter Id
+     */
+    registerEventEmitter(id: string): void {
+        this.eventEmitterMap.set(id, new vscode.EventEmitter<unknown>());
+    }
```
colin-grant-work, line 39: Should this return a `Disposable` allowing the creator of the `EventEmitter` to dispose of it and remove it from the `Map`?

vscode-trace-extension/src/utils/trace-extension-event-manager.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export interface SignalCallbackInterface {
+    signal: string,
+    callback: (data: unknown) => void
+}
+
+/**
+ *  Manages the events for the trace extension
+ */
+export class TraceExtensionEventManager {
```
colin-grant-work, line 16: This service raises some lifecycle concerns - see my other comments.

vscode-trace-extension/src/trace-explorer/opened-traces/trace-explorer-opened-traces-webview-provider.ts
```diff
@@ -16,15 +18,20 @@ const JSONBig = JSONBigConfig({
...
 
-   	constructor(
+	constructor(
 		private readonly _extensionUri: vscode.Uri,
 		private readonly _statusService: TraceServerConnectionStatusService,
-	) {}
+	) {
+	    traceExtensionEventManager.registerEventEmitter(TraceExplorerOpenedTracesViewProvider.eventEmitter_id_callback);
+	    traceExtensionEventManager.onEvent(TraceExplorerOpenedTracesViewProvider.eventEmitter_id_callback,
+	        (data: unknown) => {this.registerCallbackForSignal(data as SignalCallbackInterface);});
```
colin-grant-work, line 33: These registrations have a somewhat odd lifecycle: they live as long as the `TraceExplorerOpenedTracesViewProvider` - the life of the plugin - but the registrations in `registerCallbackForSignal` apply only to a given webview. Perhaps instead, the API should expose an event that is fired when a new webview is created, and subscribers can register whatever signal handlers they wish directly on the webview, rather than going through this indirect event -> event system?

vscode-trace-extension/src/utils/trace-extension-event-manager.ts
colin-grant-work: Let me sharpen this comment a bit. This service is not an API of this extension: nothing in this service refers specifically to functionality implemented elsewhere in the extension. The lifecycle issues, which your recent push has improved somewhat, are troubling, but fundamentally, they follow from the fact that this service has no connection to the rest of the extension. Please consider my comment about exposing instead an event when a webview is created, and perhaps a getter to expose the current webview. That is specifically relevant to the functionality implemented by the extension, and a far cleaner and easier-to-maintain solution to the only problem this service is currently solving.

bhufmann:
Thanks for the interesting discussions to work out the external API. I'll needs some time to understand the recent comments and code. I'll follow-up with you soon.

bhufmann:
Thanks again for this initiative, and also feedback of all participants. It's not an easy topic to come-up with an API that is somewhat future-proof and easy to maintain. 

Ok, it took some time to think about this and to use the proposed API. I'm able to achieve to get a callback registered for a given  signal (experimentSelectedSignal). I was also able to add additional hook inside the external API to trigger the start of the trace server. I was also able to register/deregister a listener using (onEventRegistered, onEventDisposed).

However, I have some difficulties in understanding the logic and wonder if it could be a bit simplified. 

Firstly, a third-party extension has to call `fireEvent()` to register a callback for a given signal, that is called when the openTraces view receives a message. As an extender of the vscode-trace-extension I would expect to use the vscode `EventEmitter` classes to do the same, i.e. register a listener that is called when a message is received from vscode in a given webview. The registration API for listener could be per webview by passing the webview ID. The listener receives all the messages and needs to have it's own switch/case-statement to identify the message of interest.

Since the webviews can be disposed and recreated, there needs to be an API that notifies the extending extension about that event to be able to do some action on it.

To Colin's alternative suggestion to expose the webview or a class that can get the webview, is also an alternative to the proposal. Then extending extension could subscribe to the messages themselves and have listeners to the onDidDisposed(). This would also provide the possibility to extend the vscode-trace-extension in respect to custom message handlers. 

Providing an API for the start/stop mechanism would be then separate I think and it won't be possible to use the custom message handlers as far as I can see. But still we can achieve that. 

Secondly, I noticed when registering a listener to onEventRegistered(), the initial one is missed because the open traces webview do the first fireEvent() during its initialization and at that time the extending extension is not available.

WDYT?

ngondalia:
Thanks for all of your feedback! It is tricky to design this particular API to be generic, yet simple and easy to maintain. I think for now, we can just expose the webviews like Colin suggested. Having webviews exposed gets rid of the necessity of having a generic signal handler. Also it greatly reduces the complexity of the API. For instances where we don't rely on webviews (i.e. start/stop server), we can just create specific events that the API provides a way to attach listeners to. We can address that in #159.

I will push a new patch soon.

vscode-trace-extension/src/utils/trace-extension-webview-manager.ts
```diff
@@ -0,0 +1,78 @@
...
+        if (!this.isDisposed()) {
+            return this.webviews;
+        }
+        return [];
+    }
+
+    fireWebviewCreated(_webview: vscode.WebviewView): void {
+        if (!this.isDisposed()) {
+            this.webviewCreated.fire(_webview);
+            this.addWebview(_webview);
```
colin-grant-work, line 27: It might be better to add the webview before firing the event, in case adopters have logic that expects the return from `getAllActiveWebviews` to include the just-added webview when they run their handler.

colin-grant-work (commented):
I think the overall style of this new API makes it a lot easier to follow and makes it a lot clearer what the extension expects consumers to be interested in and what consumers can expect the extension to do. 👍

colin-grant-work:
I'll leave final approval to @bhufmann, as he has more of a user / adopter's perspective in mind than I do.

bhufmann (changes requested):
Thanks for updating the proposed external API. It's much clearer to understand and to use. I have some comments though.

It would be good to have some automated test for the API.

ngondalia:
> Thanks for updating the proposed external API. It's much clearer to understand and to use. I have some comments though.
> 
> It would be good to have some automated test for the API.

I have updated the README in the latest patch with detailed instructions. The tests i can cover under a separate PR, as non UI tests will require additional configuration to setup.

bhufmann (approved):
It looks good to me now. I'll request a second committer review before merging it.

marco-miller (changes requested):
Left are only these minor comments for me; thanks!

marco-miller (approved):
LGTM too now, thanks again for this contrib.

PR #150, opened 2023-06-07
Use constants for undo, redo and updateZoom vscode messages

bhufmann:
Use constants for undo, redo and updateZoom vscode messages instead of hard-coded strings.


hriday-panchasara (approved):
LGTM, thanks!

PR #152, opened 2023-06-20
TimeRangeDataWidget CodeOSS Port

williamsyang-work:
Integrates the TimeRangeDataWidget to VSCode Trace Extension's trace explorer.

This requires changes in the theia-trace-extension repo: https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/991

### DIFF
--- a/vscode-trace-extension/src/trace-explorer/time-range/trace-explorer-time-range-data-webview-provider.ts
+++ b/vscode-trace-extension/src/trace-explorer/time-range/trace-explorer-time-range-data-webview-provider.ts
@@ -1,0 +1,154 @@
+import * as vscode from 'vscode';
+import { getTraceServerUrl } from 'vscode-trace-extension/src/utils/tspClient';
+import JSONBigConfig from 'json-bigint';
+import { Signals, signalManager } from 'traceviewer-base/lib/signals/signal-manager';
+import { TimeRangeUpdatePayload } from 'traceviewer-base/lib/signals/time-range-data-signal-payloads';
+import { TimeRangeDataMap } from 'traceviewer-react-components/lib/components/utils/time-range-data-map';
+import { Experiment } from 'tsp-typescript-client/lib/models/experiment';
+import { VSCODE_MESSAGES } from 'vscode-trace-common/lib/messages/vscode-message-manager';
+
+const JSONBig = JSONBigConfig({
+    useNativeBigInt: true,
+});
+
+export class TraceExplorerTimeRangeDataProvider implements vscode.WebviewViewProvider {
+  public static readonly viewType = 'traceExplorer.timeRangeDataView';
+  private _view: vscode.WebviewView;
+  private _disposables: vscode.Disposable[] = [];
+  private _experimentDataMap: TimeRangeDataMap;
+  constructor(private readonly _extensionUri: vscode.Uri) {
+      this._experimentDataMap = new TimeRangeDataMap();
+  }
+
+  resolveWebviewView(
+      webviewView: vscode.WebviewView,
+      _context: vscode.WebviewViewResolveContext,
+      _token: vscode.CancellationToken
+  ): void {
+      this._view = webviewView;
+      webviewView.webview.html = this._getHtmlForWebview(webviewView.webview);
+      webviewView.webview.options = {
+          // Allow scripts in the webview
+          enableScripts: true,
+          localResourceRoots: [vscode.Uri.joinPath(this._extensionUri, 'pack')],
+      };
+
+      webviewView.webview.onDidReceiveMessage(message => {
+          const command = message?.command;
+          const parsedData = message?.data ? JSONBig.parse(message.data) : undefined;
+
+          switch (command) {
+          case VSCODE_MESSAGES.REQUEST_SELECTION_RANGE_CHANGE:
+              signalManager().fireRequestSelectionRangeChange(parsedData);
+              break;
+          }
+      }, undefined, this._disposables);
+
+      webviewView.onDidChangeVisibility(() => {
+          if (this._view.visible) {
+              const data = {
+                  mapArray: Array.from(this._experimentDataMap.experimentDataMap.values()),
+                  activeData: this._experimentDataMap.activeData,
+              };
+              this._view.webview.postMessage({ command: VSCODE_MESSAGES.RESTORE_VIEW, data: JSONBig.stringify(data) });
+          }
+      });
+
+      signalManager().on(Signals.VIEW_RANGE_UPDATED, this.onViewRangeUpdated);
+      signalManager().on(Signals.SELECTION_RANGE_UPDATED, this.onSelectionRangeUpdated);
+      signalManager().on(Signals.EXPERIMENT_SELECTED, this.onExperimentSelected);
+      signalManager().on(Signals.EXPERIMENT_UPDATED, this.onExperimentUpdated);
+      signalManager().on(Signals.EXPERIMENT_CLOSED, this.onExperimentClosed);
+      signalManager().on(Signals.CLOSE_TRACEVIEWERTAB, this.onExperimentTabClosed);
+
+      webviewView.onDidDispose(_event => {
+          signalManager().off(Signals.VIEW_RANGE_UPDATED, this.onViewRangeUpdated);
+          signalManager().off(Signals.SELECTION_RANGE_UPDATED, this.onSelectionRangeUpdated);
+          signalManager().off(Signals.EXPERIMENT_SELECTED, this.onExperimentSelected);
+          signalManager().off(Signals.EXPERIMENT_UPDATED, this.onExperimentUpdated);
+          signalManager().off(Signals.EXPERIMENT_CLOSED, this.onExperimentClosed);
+          signalManager().off(Signals.CLOSE_TRACEVIEWERTAB, this.onExperimentTabClosed);
+      }, undefined, this._disposables);
+  }
+
+  private onViewRangeUpdated = (update: TimeRangeUpdatePayload) => {
+      this._view.webview.postMessage({ command: VSCODE_MESSAGES.VIEW_RANGE_UPDATED, data: JSONBig.stringify(update) });
+      this._experimentDataMap.updateViewRange(update);
+  };
+
+  private onSelectionRangeUpdated = (update: TimeRangeUpdatePayload) => {
+      this._view.webview.postMessage({ command: VSCODE_MESSAGES.SELECTION_RANGE_UPDATED, data: JSONBig.stringify(update) });
+      this._experimentDataMap.updateSelectionRange(update);
+  };
+
+  private onExperimentSelected = (experiment: Experiment | undefined) => {
+      const data = { wrapper: experiment ? JSONBig.stringify(experiment) : undefined };
+      this._view.webview.postMessage({ command: VSCODE_MESSAGES.EXPERIMENT_SELECTED, data });
+      if (experiment) {
+          this._experimentDataMap.updateAbsoluteRange(experiment);
+      }
+      this._experimentDataMap.setActiveExperiment(experiment);
+
+  };
+
+  private onExperimentUpdated = (experiment: Experiment) => {
+      const data = { wrapper: JSONBig.stringify(experiment) };
+      this._view.webview.postMessage({ command: VSCODE_MESSAGES.EXPERIMENT_UPDATED, data });
+      this._experimentDataMap.updateAbsoluteRange(experiment);
+
+  };
+
+  private onExperimentClosed = (experiment: Experiment) => {
+      const data = { wrapper: JSONBig.stringify(experiment) };
+      this._view.webview.postMessage({ command: VSCODE_MESSAGES.EXPERIMENT_CLOSED, data });
+      this._experimentDataMap.delete(experiment);
+
+  };
+
+  private onExperimentTabClosed = (experimentUUID: string) => {
+      this._view.webview.postMessage({ command: VSCODE_MESSAGES.TRACE_VIEWER_TAB_CLOSED, data: experimentUUID });
+      this._experimentDataMap.delete(experimentUUID);
+  };
+
+  /* eslint-disable max-len */
+  private _getHtmlForWebview(webview: vscode.Webview) {
+      // Get the local path to main script run in the webview, then convert it to a uri we can use in the webview.
+      const scriptUri = webview.asWebviewUri(vscode.Uri.joinPath(this._extensionUri, 'pack', 'timeRangePanel.js'));
+      const packUri = webview.asWebviewUri(vscode.Uri.joinPath(this._extensionUri, 'pack'));
+
+      // Use a nonce to only allow a specific script to be run.
+      const nonce = getNonce();
+
+      return `<!DOCTYPE html>
+			<html lang="en">
+			<head>
+				<meta charset="utf-8">
+				<meta name="viewport" content="width=device-width,initial-scale=1,shrink-to-fit=no">
+				<meta name="theme-color" content="#000000">
+				<title>React App</title>
+				<meta http-equiv="Content-Security-Policy" content="default-src 'none'; img-src vscode-resource: https:; script-src 'nonce-${nonce}' 'unsafe-eval';style-src vscode-resource: 'unsafe-inline' http: https: data:;connect-src ${getTraceServerUrl()};">
+				<base href="${packUri}/">
+			</head>
+
+			<body>
+				<noscript>You need to enable JavaScript to run this app.</noscript>
+				<div id="root"></div>
+				
+                <script nonce="${nonce}">
+					const vscode = acquireVsCodeApi();
+				</script>
+                <script nonce="${nonce}" src="${scriptUri}"></script>
+            </body>
+			</html>`;
+  }
+}
+
+function getNonce() {
+    let text = '';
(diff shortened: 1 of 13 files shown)


vscode-trace-webviews/src/trace-explorer/time-range/vscode-time-range-data-widget.tsx
```diff
@@ -0,0 +1,95 @@
...
+  componentWillUnmount(): void {
+      signalManager().off(Signals.REQUEST_SELECTION_RANGE_CHANGE, this.onRequestSelectionChange);
+  }
+
+  onRequestSelectionChange = (payload: TimeRangeUpdatePayload): void => {
+      this._signalHandler.requestSelectionRangeChange(payload);
+  };
+
+  componentDidMount(): void {
+      this._signalHandler.notifyReady();
```
colin-grant-work, line 80: Is this notification handled?

vscode-trace-webviews/src/trace-explorer/time-range/vscode-time-range-data-widget.tsx
```diff
@@ -0,0 +1,95 @@
+import * as React from 'react';
+import { Signals, signalManager } from 'traceviewer-base/lib/signals/signal-manager';
+import { ReactTimeRangeDataWidget } from 'traceviewer-react-components/lib/trace-explorer/trace-explorer-time-range-data-widget';
+import 'traceviewer-react-components/style/trace-explorer.css';
+// import { Experiment } from 'tsp-typescript-client/lib/models/experiment';
```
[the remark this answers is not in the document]
williamsyang-work: Leftover and removed.

vscode-trace-webviews/src/trace-explorer/time-range/vscode-time-range-data-widget.tsx
```diff
@@ -0,0 +1,95 @@
...
+import '../../style/react-contextify.css';
+import '../../style/trace-viewer.css';
+import JSONBigConfig from 'json-bigint';
+import { convertSignalExperiment } from 'vscode-trace-common/lib/signals/vscode-signal-converter';
+import { VSCODE_MESSAGES, VsCodeMessageManager } from 'vscode-trace-common/lib/messages/vscode-message-manager';
```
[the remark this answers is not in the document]
williamsyang-work: The signal manager can only reach within a JS GEC.  This was available for all views in the Theia IDE, but is contained inside of each webview for Code OSS.
We need to use a combination of signals and vscode-messages for the events to be heard where they are needed.  This architecture is a little convoluted but explained in the README under [Communcation Between Components](https://github.com/eclipse-cdt-cloud/vscode-trace-extension/blob/master/README.md#communication-between-components)

vscode-trace-webviews/src/trace-explorer/time-range/vscode-time-range-data-widget.tsx
williamsyang-work: I am not sure what this signal is or where it is handled.  I saw that every other webview that is in the trace-explorer fired this signal so I left it in mine.  @bhufmann, offhand, do you know if this is necessary or not?

colin-grant-work: Usually a signal like this would be used to facilitate the following sequence:
 - User opens view
 - View provider creates webview
 - When webview is ready, view provider sends it data to render

If there's no lifecycle handling that requires the plugin side to interact with the webview when it's ready (but not before), then the ready signal is redundant (but doesn't do any harm).

colin-grant-work: I understand that state restoration when the webview goes out of view is a problem for this PR. That is a potential use for this signal, as well. If the provider knows what signals to send to get the view in the right state, it can send those whenever it receives the ready signal. That way, every time the view re-mounts, the provider sends what's necessary to show the correct data in the view.

Alternatively, that could be handled on the view side using the VSCode's webview state restoration mechanisms. In that case, the webview side would save its state as the most recent, state-relevant signals it had received, and then fetch and reemit those signals on remount.

vscode-trace-webviews/src/trace-explorer/time-range/vscode-time-range-data-widget.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+  componentDidMount = (): void => {
+      this._signalHandler.notifyReady();
+      signalManager().on(Signals.REQUEST_SELECTION_RANGE_CHANGE, this.onRequestSelectionChange);
+
+  }
+
+  storeState = (): void => {
+      if (this._reactRef.current) {
+        const { state, experimentDataMap } = this._reactRef.current;
```
MatthewKhouzam, line 106: experimenterror TS2341: Property 'experimentDataMap' is private and only accessible within class 'ReactTimeRangeDataWidget'.

This was tested in Gitpod.

vscode-trace-webviews/src/trace-explorer/time-range/vscode-time-range-data-widget.tsx
```diff
@@ -0,0 +1,136 @@
...
+        });
+      }
+  }
+
+  restoreState = (): void => {
+      if (this._reactRef.current) {
+          const persistedData = vscode.getState();
+          const activeData = JSONBig.parse(persistedData.activeData);
+          const experimentDataMap: Map<string, ExperimentTimeRangeData> = new Map(JSONBig.parse(persistedData.experimentDataMap));
+          this._reactRef.current.restoreState(activeData, experimentDataMap);
```
MatthewKhouzam, line 119: error TS2339: Property 'restoreState' does not exist on type 'ReactTimeRangeDataWidget'.

vscode-trace-webviews/src/trace-viewer/vscode-trace-viewer-container.tsx
```diff
@@ -34,6 +36,13 @@ class TraceViewerContainer extends React.Component<{}, VscodeAppState>  {
 
   private _signalHandler: VsCodeMessageManager;
 
+  private onViewRangeUpdated = (payload: TimeRangeUpdatePayload): void => this._signalHandler.viewRangeUpdated(payload);
+  private onSelectionRangeUpdated = (payload: TimeRangeUpdatePayload): void => this._signalHandler.selectionRangeUpdated(payload);
+  private onRequestSelectionChange = (payload: TimeRangeUpdatePayload): void => this._signalHandler.requestSelectionRangeChange(payload);
+  private onExperimentUpdated = (payload: Experiment): void => this._signalHandler.experimentUpdated(payload);
+  // VVV Does this need to go somewhere else? VVV //
```
MatthewKhouzam, line 43: replace name with // TODO: move if need be?

vscode-trace-webviews/src/trace-explorer/time-range/vscode-time-range-data-widget.tsx
```diff
@@ -0,0 +1,136 @@
...
+    setState(data: PersistedData): void;
+    getState(): PersistedData;
+}
+
+declare const vscode: vscode;
+
+class TimeRangeDataWidget extends React.Component {
+  private _signalHandler: VsCodeMessageManager;
+  private _reactRef: React.RefObject<ReactTimeRangeDataWidget>;
+//   private restoring: boolean;
```
MatthewKhouzam, line 35: remove?

MatthewKhouzam (changes requested):
Some minor comments. I cannot get it to build locally or on gitpod, I put the error codes in the comments. Hope it helps.

vscode-trace-webviews/src/trace-explorer/time-range/vscode-time-range-data-widget.tsx
williamsyang-work: This error is happening because this needs requires changes to the theia-trace-extension: https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/991

vscode-trace-webviews/src/trace-explorer/time-range/vscode-time-range-data-widget.tsx
williamsyang-work: This error is also linked to the theia-trace-extension changes.

colin-grant-work:
Comments from demo:
 - [ ] Please remove `console.dir` fallback. In general plugins shouldn't log using the console, but should use the VSCode logging API.
 - [ ] Find and fix the BigInt conversion bug
 - [ ] With the addition of this widget, the status bar item that shows the selection is mostly redundant and should likely be removed. It appears to be generated here: 

https://github.com/eclipse-cdt-cloud/vscode-trace-extension/blob/f029c45811f50dc3874597a77a9019151cec365e/vscode-trace-extension/src/trace-viewer-panel/trace-viewer-webview-panel.ts#L178-L179

The open-ended status bar item creation implemented in the `handleStatusMessage` is a little bit troubling - the status bar shouldn't really be used for detailed communication with users, so it shouldn't be so open as to be used by anyone for anything.

bhufmann:
> Comments from demo:
> 
> * [ ]  Please remove `console.dir` fallback. In general plugins shouldn't log using the console, but should use the VSCode logging API.
> * [ ]  Find and fix the BigInt conversion bug
> * [ ]  With the addition of this widget, the status bar item that shows the selection is mostly redundant and should likely be removed. It appears to be generated here:
> 
> https://github.com/eclipse-cdt-cloud/vscode-trace-extension/blob/f029c45811f50dc3874597a77a9019151cec365e/vscode-trace-extension/src/trace-viewer-panel/trace-viewer-webview-panel.ts#L178-L179
> 
> The open-ended status bar item creation implemented in the `handleStatusMessage` is a little bit troubling - the status bar shouldn't really be used for detailed communication with users, so it shouldn't be so open as to be used by anyone for anything. Addressing the status bar item is big enough that it can be a separate ticket, rather than getting into this PR.

Please note that the status bar also show the the time delta of the cursors which is important to have so that user can measure time deltas. If we decide to remove the time values from the status bar, the delta needs to be provided differently. The Time Range widget would be a candidate for that. 

Not sure if we should remove the cursor and delta values from the status bar. That information should be visible for the user all the time and the time range widget might be closed. I agree though, that we should not use the status bar lightly and log any arbitrary message there.

colin-grant-work:
> Please note that the status bar also shows the the time delta of the cursors which is important.

Likely this should be added to the `TimeRangeDataWidget` if it's important. It looks like at the moment it only displays the absolute positions of view and selection, and not the deltas.

williamsyang-work:
The build is failing because the theia-trace-extension related changes aren't currently available.

bhufmann (changes requested):
When testing https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/991 I found the following issues with this PR:

1. When opening a trace the view range starts at 0. End time seems to be off as well (see next point)
2. The timestamp seems without offset applied
3. When Closing the last trace the values are not cleared
4. When trace server is not running the the timerange widget should not be rendered (as item properties and available views are not rendered)

The attached gif should show these issues:

time-range-widget-issues

bhufmann (changes requested):
I put some review comments and I found some issues. There first one I have already reported earlier. The time range widget should not be rendered if the trace server is not running:
TimeRangeWidgetServerDown

The second problem is intermittent. Sometimes the view range is 0 and hence the other time values are incorrect. I have not found a deterministic way to reproduce this, but when it happened I switched between different panels of multiple open traces by clicking on the panel in the middle of the window. See the picture on how it looks like:
IncorrectOffset

The third one that I noticed is when the timerange window is minimized. Changing selection, for example, when it's minimized is not applied correctly when after maximizing it. I guess there is something handling the queue.

timerange-data-incorrect-queue

The last one, is when having multiple traces open and when close one trace that is not the selected one, then the time range widget is cleared It should be only cleared only if it's the last trace that is closed.

time-range-widget-cleared-closing

vscode-trace-webviews/src/trace-explorer/time-range/vscode-time-range-data-widget.tsx
```diff
@@ -0,0 +1,142 @@
...
+      if (this._reactRef.current.isBusy) {
+          if (this._timeout) {
+              window.clearTimeout(this._timeout);
+          }
+          this._timeout = window.setTimeout(this.storeState, 100);
+          return;
+      }
+      const { activeData, experimentDataMap } = this._reactRef.current.persistedState;
+      vscode.setState({
+          activeData: JSONBig.stringify(activeData),
```
bhufmann, line 115: So, if you serialize using JSONBig, all bigint variables are serialized in a special way so that when calling JSONBig.parse() a bigint is recreated. However, if the bigint value fits into a javascript number when calling JSONBig.stringify() it won't serialize it that special way and when calling JSON.big.parse() the corresponding variable is NOT a bigint. This can lead to problems later on.

So, this is the reason why the tsp-typescript-client has special code to account for that and the vscode-trace-extension has some special code for serializing/deserializing Experiment. See vscode-signal-converter.ts.

So, you will need take care of this when doing JSONBig.parse(),

BTW, we plan to streamline both vscode-signal-converter.ts and tsp-typescript-client for the bigint handling (not scope of this PR).

MatthewKhouzam:
I am getting an endless loop on this, it does not populate with traces or experiments.

Tested with:

236c8d01395ebf2a682e651b94d666e5b09c5309 -> vscode
0533d18f609822b3d24c4d562745639d8eef4213 -> theia side

In VSCodium 1.75. and glorious Theia 1.39

MatthewKhouzam:
Some issues!

1- needs to have same tsp-client-version as #991 from theia-trace-extension

Long text should turn red when it's impossible to be correct (or larger than end)
https://github.com/eclipse-cdt-cloud/vscode-trace-extension/assets/419535/4f194e01-9a9c-4c17-8916-8c23f19b9d0a

It's impossible to have an empty text box for "pasting in" a good timestamp (from excel)
https://github.com/eclipse-cdt-cloud/vscode-trace-extension/assets/419535/665370d8-d04a-4689-81f9-7474ca8d4794

When the time is invalid, it should be re-validated. 
https://github.com/eclipse-cdt-cloud/vscode-trace-extension/assets/419535/d4746d03-5569-4faa-a1b3-e096980a4d5b

williamsyang-work:
Notes: 

>needs to have same tsp-client-version as #991 from theia-trace-extension
This is fixed in the most recent commit.

> Long text should turn red when it's impossible to be correct (or larger than end) https://github.com/eclipse-cdt-cloud/vscode-trace-extension/assets/419535/4f194e01-9a9c-4c17-8916-8c23f19b9d0a
The color of the input-box numbers wasn't originally red.  It was changed in [this commit in theia-trace-extension].(https://github.com/eclipse-cdt-cloud/theia-trace-extension/commit/e1842b7087b376befa485718c239b709099867a6)

This has been fixed by adding the variable `--trace-viewer-input-foreground` to vscode's css.


> It's impossible to have an empty text box for "pasting in" a good timestamp (from excel) https://github.com/eclipse-cdt-cloud/vscode-trace-extension/assets/419535/665370d8-d04a-4689-81f9-7474ca8d4794
[This has been fixed in the other repository, along with some other buggy behaviors.](https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/991/commits/99c3b151b21e3e6e32cc8b109ead9736d02b51ea)


> When the time is invalid, it should be re-validated. https://github.com/eclipse-cdt-cloud/vscode-trace-extension/assets/419535/d4746d03-5569-4faa-a1b3-e096980a4d5b

Data validation has always been on submit, not on change.  I think the change to the color of the numbers made the UX jarring.  I'm not going to change anything since fixing the font-color seemed to correct the experience.

vscode-trace-extension/src/trace-explorer/time-range/trace-explorer-time-range-data-webview-provider.ts
```diff
@@ -0,0 +1,154 @@
...
+      signalManager().on(Signals.EXPERIMENT_CLOSED, this.onExperimentClosed);
+      signalManager().on(Signals.CLOSE_TRACEVIEWERTAB, this.onExperimentTabClosed);
+
+      webviewView.onDidDispose(_event => {
+          signalManager().off(Signals.VIEW_RANGE_UPDATED, this.onViewRangeUpdated);
+          signalManager().off(Signals.SELECTION_RANGE_UPDATED, this.onSelectionRangeUpdated);
+          signalManager().off(Signals.EXPERIMENT_SELECTED, this.onExperimentSelected);
+          signalManager().off(Signals.EXPERIMENT_UPDATED, this.onExperimentUpdated);
+          signalManager().off(Signals.EXPERIMENT_CLOSED, this.onExperimentClosed);
+          signalManager().on(Signals.CLOSE_TRACEVIEWERTAB, this.onExperimentTabClosed);
```
MatthewKhouzam, line 70: Make sure this is correct

yarn.lock
MatthewKhouzam, line 1: Update this... but then this is all good to go! great job!

MatthewKhouzam (approved):
Aproving assuming you change the yarn.lock and rebase. Needs a second review though.

MatthewKhouzam:
Looks great, get the build working! :)

hriday-panchasara (approved):
Works great, thanks!

marco-miller:
@williamsyang-work is looking into marking the still open conversation threads as Resolved or so. Once he is done with that, I'm in turn planning to merge this PR. Given that we have our two approvals already. Thanks.

PatrickTasse (approved):

PR #155, opened 2023-06-27
Automatic start trace server when user requests to open a trace

### LINKED ISSUE #15: Migrate start and stop of trace server using the backend

bhufmann:
This change uses the vscode-trace-server commands to start a trace server and when the server is started it will perform the pending open command.

This change relies on the [vscode-trace-server](https://github.com/eclipse-cdt-cloud/vscode-trace-server) extension.

In a future release an API can be introduced so that an extension can register the start/stop commands and also the started notification. This future enhancement would decouple the vscode-trace-extension with the [vscode-trace-server](https://github.com/eclipse-cdt-cloud/vscode-trace-server) implementation.

Fixes #15

### DIFF
--- a/vscode-trace-extension/src/extension.ts
+++ b/vscode-trace-extension/src/extension.ts
@@ -6,12 +6,14 @@ import { TraceExplorerAvailableViewsProvider } from './trace-explorer/available-
 import { TraceExplorerOpenedTracesViewProvider } from './trace-explorer/opened-traces/trace-explorer-opened-traces-webview-provider';
 import { fileHandler, openOverviewHandler, resetZoomHandler, undoRedoHandler, zoomHandler, keyboardShortcutsHandler } from './trace-explorer/trace-tree';
 import { TraceServerConnectionStatusService } from './utils/trace-server-status';
-import { updateTspClient } from './utils/tspClient';
+import { getTspClientUrl, updateTspClient } from './utils/tspClient';
 import { TraceExtensionLogger } from './utils/trace-extension-logger';
 import { VSCODE_MESSAGES } from 'vscode-trace-common/lib/messages/vscode-message-manager';
 import { TraceViewerPanel } from './trace-viewer-panel/trace-viewer-webview-panel';
+import { TspClientProvider } from 'vscode-trace-common/lib/client/tsp-client-provider-impl';
 
 export let traceLogger: TraceExtensionLogger;
+const tspClientProvider = new TspClientProvider(getTspClientUrl(), undefined);
 
 export function activate(context: vscode.ExtensionContext): void {
     traceLogger = new TraceExtensionLogger('Trace Extension');
@@ -42,8 +44,11 @@ export function activate(context: vscode.ExtensionContext): void {
     const analysisProvider = new AnalysisProvider();
     // TODO: For now, a different command opens traces from file explorer. Remove when we have a proper trace finder
     const fileOpenHandler = fileHandler(analysisProvider);
-    context.subscriptions.push(vscode.commands.registerCommand('traces.openTraceFile', file => {
-        fileOpenHandler(context, file);
+    context.subscriptions.push(vscode.commands.registerCommand('traces.openTraceFile', async file => {
+        await startTraceServerIfAvailable();
+        if (await isUp()) {
+            fileOpenHandler(context, file);
+        }
     }));
 
     // Listening to configuration change for the trace server URL
@@ -88,8 +93,11 @@ export function activate(context: vscode.ExtensionContext): void {
         TraceViewerPanel.showMarkersFilterOnCurrent();
     }));
 
-    context.subscriptions.push(vscode.commands.registerCommand('openedTraces.openTraceFolder', () => {
-        fileOpenHandler(context, undefined);
+    context.subscriptions.push(vscode.commands.registerCommand('openedTraces.openTraceFolder', async () => {
+        await startTraceServerIfAvailable();
+        if (await isUp()) {
+            fileOpenHandler(context, undefined);
+        }
     }));
 
     context.subscriptions.push(vscode.commands.registerCommand('traceViewer.shortcuts', () => {
@@ -113,4 +121,19 @@ export function activate(context: vscode.ExtensionContext): void {
 
 export function deactivate(): void {
     traceLogger.disposeChannel();
+}
+
+async function startTraceServerIfAvailable(): Promise<void> {
+    const extensionId = 'vscode-trace-server';
+    const traceServerExtension = vscode.extensions.getExtension('tracecompass-community.' + extensionId);
+    if (!traceServerExtension || await isUp()) {
+        return;
+    }
+    await vscode.commands.executeCommand(extensionId + '.start-if-stopped');
+}
+
+async function isUp() {
+    const health = await tspClientProvider.getTspClient().checkHealth();
+    const status = health.getModel()?.status;
+    return health.isOk() && status === 'UP';
 }


bhufmann:
Replaces #142 which this one is inspired by.

vscode-trace-extension/src/extension.ts
```diff
@@ -6,15 +6,24 @@ import { TraceExplorerAvailableViewsProvider } from './trace-explorer/available-
...
 import { fileHandler, openOverviewHandler, resetZoomHandler, undoRedoHandler, zoomHandler, keyboardShortcutsHandler } from './trace-explorer/trace-tree';
 import { TraceServerConnectionStatusService } from './utils/trace-server-status';
-import { updateTspClient } from './utils/tspClient';
+import { updateTspClient, getTspClientUrl } from './utils/tspClient';
 import { TraceExtensionLogger } from './utils/trace-extension-logger';
 import { VSCODE_MESSAGES } from 'vscode-trace-common/lib/messages/vscode-message-manager';
+import { TspClientProvider } from 'vscode-trace-common/lib/client/tsp-client-provider-impl';
 
 export let traceLogger: TraceExtensionLogger;
+let tspClientProvider: TspClientProvider;
```
colin-grant-work, line 15: As far as I can tell, this could just be initialized immediately, since it doesn't depend on anything that's passed into the `activate` function, and that would save some checks for whether it's defined.

vscode-trace-extension/src/extension.ts
```diff
@@ -102,3 +124,38 @@ export function activate(context: vscode.ExtensionContext): void {
...
+    return true;
+}
+
+async function openTrace(analysisProvider: AnalysisProvider, content: PendingOpen) {
+    pendingOpen = content;
+    const fileOpenHandler = fileHandler(analysisProvider);
+    const startPromise = startTraceServerIfAvailable();
+    startPromise.then(() => {
+        if (pendingOpen === undefined) {
+            fileOpenHandler(content.context, content.file);
```
colin-grant-work, line 153: The code handling `pendingOpen` is a bit odd. The only time there's a positive reference is in the handler for `serverStatus.started`, where something whose opening is pending will be opened. But the only place that `pendingOpen` is set is here in the `openTrace` function, where we insist on opening whatever's passed in, but only(?) if someone has set `pendingOpen` to `undefined` in the meantime. It basically seems to be aimed at two scenarios: if the server actually requires starting, then the server-starting plugin will invoke `serverStatus.started` when it starts and then the trace will be opened in that handler; alternatively, if the server is already up, then `pendingOpen` gets set to `undefined`, and `openTrace` handles the actual opening, but a piece of global state seems like a fragile way to handle those options.

Some possible solutions:
 - Ideally, the handler for `vscode-trace-server.start-if-stopped` wouldn't return until it had started the server or knew it couldn't. Then the code would just look like:
 ```
await startTraceServerIfAvailable();
if (await isUp()) {
    fileOpenHandler(content.context, content.file);
}
```
- Rather than adding `pendingOpen` to the global context, you could replace the command handler for `serverStatus.started` with one that performed the opening. That would have the advantage of at least keeping the variable local.

File: vscode-trace-extension/src/extension.ts [hunk 3]
```diff
@@ -102,3 +124,38 @@ export function activate(context: vscode.ExtensionContext): void {
...
 }
+
+async function startTraceServerIfAvailable(): Promise<void>{
+    const traceServerExtension = vscode.extensions.getExtension('tracecompass-community.vscode-trace-server');
+    if (!traceServerExtension || await isUp()) {
+        pendingOpen = undefined;
+        return;
+    }
+    const extensionId = 'vscode-trace-server';
+    vscode.commands.executeCommand(extensionId + '.start-if-stopped');
```
colin-grant-work, line 135: In the context of `openTrace` below, it doesn't seem guaranteed that `pendingOpen` will either be consumed or be set to `undefined`. If the command is executed but neither returned nor awaited, then in `openTrace`, we don't get into the `catch` block that would set `pendingOpen` to `undefined`, but it isn't guaranteed that the other plugin will eventually call `serverStatus.started` (e.g. if the server fails to start), so `pendingOpen` would remain set to the old value.

vscode-trace-extension/src/extension.ts
bhufmann: Thank you very much for the review!

> The code handling `pendingOpen` is a bit odd. The only time there's a positive reference is in the handler for `serverStatus.started`, where something whose opening is pending will be opened. But the only place that `pendingOpen` is set is here in the `openTrace` function, where we insist on opening whatever's passed in, but only(?) if someone has set `pendingOpen` to `undefined` in the meantime. It basically seems to be aimed at two scenarios: if the server actually requires starting, then the server-starting plugin will invoke `serverStatus.started` when it starts and then the trace will be opened in that handler; alternatively, if the server is already up, then `pendingOpen` gets set to `undefined`, and `openTrace` handles the actual opening, but a piece of global state seems like a fragile way to handle those options.
> 
> Some possible solutions:
> 
>     * Ideally, the handler for `vscode-trace-server.start-if-stopped` wouldn't return until it had started the server or knew it couldn't. Then the code would just look like:
> 
> 
> ```
> await startTraceServerIfAvailable();
> if (await isUp()) {
>    fileOpenHandler(content.context, content.file);
> }
> ```
That was my initial idea as well. However, I was not able to make it work that `startTraceServerIfAvailable` waits until the server is actually running. The corresponding call to `vscode.executeCommand(...)` always returned right away and the start procedure was still ongoing while fileOpenHandler() was called. This let to issues when opening a trace from the file explorer (without the open dialog). Then I thought about using the `serverStatus.started` notification as trigger for calling `fileOpenHandler()`. It works, but the implementation was not clean and I was not really happy about it either. 

I just revisited the implementation of the `vscode-trace-server.start-if-stopped` command handler and I noticed that the command handler was not async nor awaiting properly, which caused the issues I had. I'll upload a patch in the `vscode-trace-server` repository to fix this. After that I'll update this PR. The solution will be simpler and cleaner afterwards.

> 
>     * Rather than adding `pendingOpen` to the global context, you could replace the command handler for `serverStatus.started` with one that performed the opening. That would have the advantage of at least keeping the variable local.

With my planed updates this won't be necessary anymore.

vscode-trace-extension/src/extension.ts
bhufmann: With my planed updates this won't be necessary anymore.

vscode-trace-extension/src/extension.ts
bhufmann: I'll check this.

vscode-trace-extension/src/extension.ts
```diff
@@ -6,11 +6,13 @@ import { TraceExplorerAvailableViewsProvider } from './trace-explorer/available-
...
 import { fileHandler, openOverviewHandler, resetZoomHandler, undoRedoHandler, zoomHandler, keyboardShortcutsHandler } from './trace-explorer/trace-tree';
 import { TraceServerConnectionStatusService } from './utils/trace-server-status';
-import { updateTspClient } from './utils/tspClient';
+import { getTspClientUrl, updateTspClient } from './utils/tspClient';
 import { TraceExtensionLogger } from './utils/trace-extension-logger';
 import { VSCODE_MESSAGES } from 'vscode-trace-common/lib/messages/vscode-message-manager';
+import { TspClientProvider } from 'vscode-trace-common/lib/client/tsp-client-provider-impl';
 
 export let traceLogger: TraceExtensionLogger;
+const tspClientProvider: TspClientProvider = new TspClientProvider(getTspClientUrl(), undefined);
```
colin-grant-work, line 15:
```suggestion
const tspClientProvider = new TspClientProvider(getTspClientUrl(), undefined);
```

vscode-trace-extension/src/extension.ts
```diff
@@ -102,3 +110,21 @@ export function activate(context: vscode.ExtensionContext): void {
...
+    const extensionId = 'vscode-trace-server';
+    const traceServerExtension = vscode.extensions.getExtension('tracecompass-community.' + extensionId);
+    if (!traceServerExtension || await isUp()) {
+        return;
+    }
+    await vscode.commands.executeCommand(extensionId + '.start-if-stopped');
+}
+
+async function isUp() {
+    if (tspClientProvider) {
```
colin-grant-work, line 124: Shouldn't need this check now, as the `tspClientProvider` is initialized unconditionally and never reassigned.

colin-grant-work (commented):
The code looks pretty good to me. Perhaps @hriday-panchasara could give it a quick functional test to catch any subtleties I've missed.

vscode-trace-extension/src/extension.ts
bhufmann: I can change that.

vscode-trace-extension/src/extension.ts
bhufmann: I can change that too.

hriday-panchasara (approved):
Works as expected, thanks for the contribution.

bhufmann:
I'll open an new issue tracker for introducing an external API so that an extension can register the start/stop commands and also the started notification. This future enhancement would decouple the vscode-trace-extension with the [vscode-trace-server](https://github.com/eclipse-cdt-cloud/vscode-trace-server) implementation.

PR #158, opened 2023-07-04
Propagate new Trace Server url upon change

### LINKED ISSUE #10: Migrate handling of preferences for trace server
The vscode plugin has defined preferences, but the handling of updates needs to be migrated.

hoangphamEclipse:
This commit allow the VSCode extension to listen to the onDidChangeConfiguration event provided by VSCode and propagate the new URL to components. The new URL is sent as a message to the front end, then the TraceServerUrlProvider will run handlers that was registered to the TSPClientProvider.

Partially fix #10.


vscode-trace-extension/src/extension.ts
```diff
@@ -51,6 +51,15 @@ export function activate(context: vscode.ExtensionContext): void {
...
         }
+
+        if (e.affectsConfiguration('trace-compass.traceserver.url')) {
+            // Signal the change to the `Opened traces` and `Available views` webview
+            tracesProvider.updateTraceServerUrl();
+            myAnalysisProvider.updateTraceServerUrl();
+
+            // Signal the change to all trace panels
+            TraceViewerPanel.updateTraceServerUrl();
+        }
```
colin-grant-work, line 62: It looks odd that the `updateTraceServerUrl` methods don't accept the new value. How do we guarantee that they all refer to the same value in the end?

vscode-trace-webviews/src/common/trace-server-url-provider.ts
colin-grant-work, line 1: This file's contents are identical to `vscode-trace-common/src/server/trace-server-url-provider.ts`

vscode-trace-extension/src/trace-viewer-panel/trace-viewer-webview-panel.ts
```diff
@@ -121,6 +121,14 @@ export class TraceViewerPanel {
 	    }
 	}
 
+	public static updateTraceServerUrl(): void {
+	    for (const panelName in TraceViewerPanel.activePanels) {
+	        if (TraceViewerPanel.activePanels[panelName]) {
+	            TraceViewerPanel.activePanels[panelName]?._panel.webview.postMessage({command: VSCODE_MESSAGES.TRACE_SERVER_URL_CHANGED, data: getTraceServerUrl()});
+	        }
+	    }
```
colin-grant-work, line 129:
```suggestion
	    Object.values(TraceViewerPanel.activePanels)
			.forEach(trace => trace?._panel.webview.postMessage({ command: VSCODE_MESSAGES.TRACE_SERVER_URL_CHANGED, data: getTraceServerUrl() }));
```

vscode-trace-webviews/src/common/trace-server-url-provider.ts
hoangphamEclipse: This is a mistake while rebasing. I will remove it.

vscode-trace-extension/src/extension.ts
hoangphamEclipse: We have a function called `getTraceServerUrl()` in the `vscode-trace-extension\src\utils\tspClient.ts` that gets the latest url and perform some additional formatting. I use this function inside each `updateTraceServerUrl()` function to get the url.

hoangphamEclipse:
@colin-grant-work thank you for your comments. I have updated my PR.

vscode-trace-extension/src/extension.ts
colin-grant-work: Is there a reason to leave the separate implementations to call that function, rather than calling it once here and passing the value in?

hoangphamEclipse: There is no particular reasons. For clarity, I think your suggestion is easier to read/maintain. I will make the adjustment.

vscode-trace-webviews/src/trace-explorer/opened-traces/vscode-trace-explorer-opened-traces-widget.tsx
```diff
@@ -57,20 +59,24 @@ class TraceExplorerOpenedTraces extends React.Component<{}, OpenedTracesAppState
...
+                  if (this.state.tspClientProvider) {
+                      this.state.tspClientProvider.addTspClientChangeListener(() => {
+                          if (this.state.tspClientProvider) {
+                              this._experimentManager = this.state.tspClientProvider.getExperimentManager();
+                          }
+                      });
+                  }
+              };
+
+              this.setState({ tspClientProvider: tspClientProvider }, callback);
```
colin-grant-work, line 79: This looks like it could be a memory leak. What happens when a new `TspClientProvider` is created? The old one will still have the callback that was added to it, and that could cause it to be retained even though we're done with it.

The same is perhaps true of other cases where new `TspClientProviders` are created, but it's less clear who's subscribing to their events.

vscode-trace-extension/src/extension.ts
```diff
@@ -56,6 +57,17 @@ export function activate(context: vscode.ExtensionContext): void {
         if (e.affectsConfiguration('trace-compass.traceserver.url') || e.affectsConfiguration('trace-compass.traceserver.apiPath')) {
             updateTspClient();
         }
+
+        if (e.affectsConfiguration('trace-compass.traceserver.url')) {
```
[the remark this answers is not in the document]
bhufmann: These preferences are a front-end settings and part of the `vscode-trace-extension`. The Front-end is connecting towards the trace server using HTTP and it needs to know the URL (incl port) and api path. This is independent to the `vscode-trace-server` extension which is just there to start and stop the trace server. Of course, in both extensions the configuration (e.g. port) needs to match to have successful TSP calls.

vscode-trace-webviews/src/trace-explorer/opened-traces/vscode-trace-explorer-opened-traces-widget.tsx
hoangphamEclipse: Please correct me if I am wrong. If you are talking about the callback at line 69, then it will only be run when a new tspClientProvider is set. If you are talking about the listener on line 71, I was just trying to keep things the same as the existing code here and in the `theia-trace-extension`. Perhaps it is an issue that is outside of the scope for this PR. If you think so, please let me know and I can create an issue in GitHub for that.

PatrickTasse: The way I understand the code, when a new TspClientProvider is created, a listener is added to itself, so no one else has a reference to it. When the old TspClientProvider is replaced and there are no more references to it, it should be discarded along with its listener.

The usual error pattern for such leaks is when the listener is added to an external object that remains when the object that added the listener is disposed...

It appears though that if a zombie TspClientProvider receives a client change event, its listener will execute, but it will set the experiment manager to the 'new' TspClientProvider's experiment manager (in addition to that new TspClientProvider also doing the same thing). So it could double set to the same value, maybe harmless? To avoid that there could be a check that the current state's tspClientProvider is the same one that created that particular listener. But I'm not sure if it's possible for the non-current TspClientProvider to receive a client change event.

vscode-trace-extension/src/extension.ts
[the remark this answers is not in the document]
colin-grant-work: Sounds fair - so at the moment, it's a matter of redundancy, rather than implicit dependency.

vscode-trace-webviews/src/trace-explorer/opened-traces/vscode-trace-explorer-opened-traces-widget.tsx
colin-grant-work: It seem unlikely that it will result in a memory leak in practice.

colin-grant-work (approved):
I have no more objections, but it would be good if another committer could give it a functional test.

vscode-trace-extension/src/extension.ts
```diff
@@ -59,6 +60,17 @@ export function activate(context: vscode.ExtensionContext): ExternalAPI {
...
+            const newUrl = getTraceServerUrl();
```
colin-grant-work, line 65: Investigating remote deployments of this plugin, I've been doing some checking on VSCode utilities for Remote deployments and found the `asExternalUri` function in the [`env`](https://code.visualstudio.com/api/references/vscode-api#env) namespace. Basically, in a desktop deployment, that function does nothing, because `external` is the same as local, but in a remote deployment, that function transforms the URI into a form that the client can use to access a client running on the host and also does any port forwarding, if the application is set up for that. You can see it in action in the ['Live Preview'](https://github.com/microsoft/vscode-livepreview/blob/d37c5d583a53db5add2ac476c3d30c4b3c13da95/src/connectionInfo/connection.ts#L74-L75) plugin, which, like the Trace Extension, opens up a server on the remote machine on certain ports. With that in mind, perhaps the output of `getTraceServerUrl` should be passed through that utility to ensure that it's accessible.

The caveat that I can see is that if a user is using a VSCode deployment that does not implement the necessary infrastructure, it might be desirable to let the user specify the exact shape of the URL they want the views to use so that the user can set up access to the service manually.

PatrickTasse (approved):

vscode-trace-extension/src/extension.ts
hoangphamEclipse: Hi @colin-grant-work, thank you for the investigation. Are you OK to merge this PR and discuss this later if necessary?

PR #164, opened 2023-07-12
Update experiment when re-opening an existing experiment on the server

### LINKED ISSUE #165: Experiment start and end time not available when re-opening experiment after trace start
The Experiment models in the open traces view don't have the start and end time when getting the experiments the first time after opening the experiment. When re-opening such experiment that model with missing start/end time is propagated to other views and might cause issues later on. 

Make sure that the experiment is updated on the back-end before propagating the experiment during re-opening. 

This can only be reproduced in the debugger atm or with #152.

bhufmann:
Before, when re-opening an experiment right after the trace server started, the Experiment didn't have the start and end time set and the Experiment was still CLOSED.

This patch triggers a fetchExperiment call to open the experiment and to get the current start time and end time.

This patch is needed to fix an issue found with #152.

Fixes #165


marco-miller (approved):
LGTM too, based on some local debugging of this patch; thanks.

PR #167, opened 2023-07-14
Use webview.asWebviewUri for generating script paths.

williamsyang-work:
Makes all URI builders utilize webview.asWebviewUri method.  This ensures that the files can be served to remote clients.


bhufmann (approved):
Looks good to me. Thanks!

marco-miller (approved):
LGTM too; thanks.

PR #172, opened 2023-07-20
Add pack dir to .vscodeignore  file

williamsyang-work:
Removes the pack directory from the extension build process.


williamsyang-work:
I had an .vsix file that was about 30GB and took 20 minutes to unzip!  Extension .vsix size is now around 300kB.

PR #173, opened 2023-08-05
Use asExternalUri For Server Url

williamsyang-work:
~~Requires: https://github.com/eclipse-cdt-cloud/vscode-trace-server/pull/11~~

This allows for automatic port forwarding of trace-server api calls when using remote.


williamsyang-work:
Sometimes after the port is forwarded, the explorer webviews do not refresh to load/show the new data from the trace-server.  There may be other UX flow issues with these changes.  I can address them if needed but thought the bugs could be documented and addressed as a separate issue.

williamsyang-work:
I tested this by building the .vsix files and installing them onto vscode.  But using the Extension Development Host also seems to work.  Just make sure that the port is not being forwarded on another open editor.

vscode-trace-extension/src/extension.ts
```diff
@@ -50,11 +50,23 @@ export function activate(context: vscode.ExtensionContext): ExternalAPI {
         }
     }));
 
+    const updateUris = async (url?: string): Promise<void> => {
+        url = url ? url : getTraceServerUrl();
```
marco-miller, line 54: `url = url ?? getTraceServerUrl();` (taken from SonarLint).

vscode-trace-extension/src/extension.ts
```diff
@@ -50,11 +50,23 @@ export function activate(context: vscode.ExtensionContext): ExternalAPI {
...
+        TraceViewerPanel.updateTraceServerUrl(extUriString);
+    };
+
     const analysisProvider = new AnalysisProvider();
     // TODO: For now, a different command opens traces from file explorer. Remove when we have a proper trace finder
     const fileOpenHandler = fileHandler(analysisProvider);
     context.subscriptions.push(vscode.commands.registerCommand('traces.openTraceFile', async file => {
-        await startTraceServerIfAvailable();
+        const url = await startTraceServerIfAvailable();
+        await updateUris(url);
```
marco-miller, line 69: The `url` returned above has no trailing slash, AFAIK. As opposed to what this method call expects based on line 54. Not sure if this can be an issue.

vscode-trace-extension/src/extension.ts
```diff
@@ -114,7 +119,8 @@ export function activate(context: vscode.ExtensionContext): ExternalAPI {
     }));
 
     context.subscriptions.push(vscode.commands.registerCommand('openedTraces.openTraceFolder', async () => {
-        await startTraceServerIfAvailable();
+        const url = await startTraceServerIfAvailable();
+        await updateUris(url);
```
marco-miller, line 123: Per above.

marco-miller (changes requested):
Thanks!

vscode-trace-extension/src/extension.ts
williamsyang-work: The code is working as expected.  I'm not sure how exactly this is handled in the code, but it is getting passed through `vscode.Uri.parse()` and my guess is that will add or remove the trailing '/' to what vscode likes.

marco-miller (approved):
LGTM, yet letting @bhufmann cover the final assessment.

vscode-trace-extension/src/extension.ts
```diff
@@ -50,11 +50,23 @@ export function activate(context: vscode.ExtensionContext): ExternalAPI {
...
+        url = url ?? getTraceServerUrl();
+        const baseUri = vscode.Uri.parse(url);
+        const extUri = await vscode.env.asExternalUri(baseUri);
+        const extUriString = extUri.toString();
+
+        tracesProvider.updateTraceServerUrl(extUriString);
+        myAnalysisProvider.updateTraceServerUrl(extUriString);
+        TraceViewerPanel.updateTraceServerUrl(extUriString);
+    };
+
```
bhufmann, line 63: @williamsyang-work a question, what if a user executes the command `Trace Server: Start (if stopped)` manually using the command palette, shouildn't `updateUris()` be called as well? Maybe when command `serverStatus.started` is executed? Please let me know.

williamsyang-work: @bhufmann I tested starting and stopping the trace server and it worked as-is for me.
My understanding is: as long as server url remains the same we don't need to update it.  The process on that port can stop, start, restart, etc.

bhufmann: @williamsyang-work When connecting to the remote host and executing the command `Trace Server: Start (if stopped)` then the server is started, but the port forwarding is not working because `updateUris()` and hence`vscode.asExternalUrl` is not executed. Start and stop is still working in this case. I tested that case with a remote running in a virtual machine and I saw it not working.

However, when using the open trace command that starts the server and `updateUris()` is executed will enable port forwarding for the manual start command.

I think we need to address the manual start case as well.

bhufmann:
@williamsyang-work I noticed that when I close the remote vscode window, the server is not automatically stopped and keeps running afterwards. I had to go to a shell and kill the process. Do you see this behaviour as well?

vscode-trace-extension/src/extension.ts
williamsyang-work: Per our discussion `updateUris()` is now called in the `serverStatus.started` command.  This updates the webview and starts port forwarding.  

There seems to be another bug where it shows the server's data accurately, but then changes back to 'No Traces Open: Open Trace' state after some time.

williamsyang-work:
> @williamsyang-work I noticed that when I close the remote vscode window, the server is not automatically stopped and keeps running afterwards. I had to go to a shell and kill the process. Do you see this behaviour as well?

Yes, this is the behavior I get as well.

williamsyang-work:
The `url` that's returned from `server.start-if-stopped` command.  Therefore, there is no longer a need for the changes in vscode-trace-server repo.

bhufmann:
> > @williamsyang-work I noticed that when I close the remote vscode window, the server is not automatically stopped and keeps running afterwards. I had to go to a shell and kill the process. Do you see this behaviour as well?
> 
> Yes, this is the behavior I get as well.

Thanks for confirming. I did some investigation and with some logging (to file) I was able to confirm, that the `deactivate()` method of the `vscode-trace-server` is being called. However, the call the `tree-kill(pid)` has no effect. More investigation needed to fix stopping the trace server on exit.

bhufmann:
> > > @williamsyang-work I noticed that when I close the remote vscode window, the server is not automatically stopped and keeps running afterwards. I had to go to a shell and kill the process. Do you see this behaviour as well?
> > 
> > 
> > Yes, this is the behavior I get as well.
> 
> Thanks for confirming. I did some investigation and with some logging (to file) I was able to confirm, that the `deactivate()` method of the `vscode-trace-server` is being called. However, the call to `tree-kill(pid)` has no effect. More investigation needed to fix stopping the trace server on exit.

I did some more digging to figure out what's going on. I think what's happening is that VsCode exits before the `treeKill()` command is executed. `treeKill()` spawns external processes to get all the children pids from the parent pid of the trace server launcher. It reads the stdout of the command `ps -o pid --no-headers --ppid $ppid` and then uses the collected list of pids to call process.kill on it. This is done in a callback, but the parent VsCode application already exited and the hence the no process is killed. I was able to make it work by adding some async/wait/timeout call to delay the execution. 

Please note that this can also happen in the local case. I guess, that the shutdown of a local VsCode application takes longer than closing the remote processes.

PR #174, opened 2023-08-10
Added external API to contribute start/stop trace server implementation

### LINKED ISSUE #159: Provide external API to contribute start/stop trace server implementation
This issue tracker is for introducing an external API so that an extension can register start/stop commands and also get start notification. This future enhancement would decouple the vscode-trace-extension with the [vscode-trace-server](https://github.com/eclipse-cdt-cloud/vscode-trace-server) implementation. This update will replace current implementation provided with #155.

See also #149 about the external API discussions.

Initiated based on [1] https://www.eclipse.org/lists/tracecompass-dev/msg01875.html.

AlexDoanBB:
Changes made:

    Added new external api for trace server startup and shutdown

The new API allows adopting extension to register server startup handler and shutdown handlers, as well as the list of trace file extensions that the adopting extension would be prioritized for.

Fixes #159

Signed-off-by: Alex Doan vadoan@blackberry.com

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,2 +1,2 @@
 # Code formatted w Prettier
-e16c0c01dc0366ca57693c2d9fe00cb8834dbc8b
+d9243b91aa114f37964a387f063c29f2c4c64fda


vscode-trace-extension/src/utils/trace-server-manager.ts
```diff
@@ -0,0 +1,86 @@
...
+                await wildcardHandler[1][0]();
+                this.stopHandlerToUse = wildcardHandler[1][1];
+                return;
+            }
+            // execute handler for extension with no file extension listed if found
+            const noExtHandlers = this.fileExtsHandlers.get([]);
+            noExtHandlers?.[0];
+            this.stopHandlerToUse = noExtHandlers?.[1];
+        }
+    }
```
marco-miller, line 68: - I'd have a concern about how to formally test this method above.
- And on a similar note, should we consider some PR to make the [vscode-trace-server](https://github.com/eclipse-cdt-cloud/vscode-trace-server) extension register itself as handlers by default (if living beside this very extension)?
- These questions would also be for @bhufmann; thanks.

vscode-trace-extension/src/extension.ts
```diff
@@ -15,9 +15,11 @@ import { VSCODE_MESSAGES } from 'vscode-trace-common/lib/messages/vscode-message
 import { TraceViewerPanel } from './trace-viewer-panel/trace-viewer-webview-panel';
```
marco-miller, line 15: This comment is about the commit message: the `Signed-off-by` is missing angle brackets around the email. You may check examples in the git log if needed.

marco-miller (dismissed):
Thanks for this contribution.

marco-miller:
Letting @bhufmann and @MatthewKhouzam review this work, as I recently changed focus (at work).

bhufmann:
@AlexDoanBB, FYI, we just merged a PR to use prettier for code formatting (see #176). It will fail the build if not formatted properly. Please you change rebase to the latest in the repository and apply prettier settings as described in this PR #176.

vscode-trace-extension/src/utils/trace-server-manager.ts
```diff
@@ -0,0 +1,93 @@
...
+            }
+        }
+    }
+
+    /**
+     * Look for appropriate start handler based on the file extension and execute it
+     *
+     * @param fileExt file extension from the trace file url
+     */
+    async startServer(fileExt: string): Promise<void> {
```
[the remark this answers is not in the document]
AlexDoanBB: that's a good idea. I don't see any issue with this implementation, I'll switch to it

AlexDoanBB:
@PatrickTasse please take a look and let me know if the changes made are according to your suggestion. Thanks!

vscode-trace-extension/src/utils/trace-server-manager.ts
AlexDoanBB: > * I'd have a concern about how to formally test this method above.
* We might be able to test the API by mocking an adopting extension and mock opening trace files

> * And on a similar note, should we consider some PR to make the [vscode-trace-server](https://github.com/eclipse-cdt-cloud/vscode-trace-server) extension register itself as handlers by default (if living beside this very extension)?
*  [vscode-trace-server](https://github.com/eclipse-cdt-cloud/vscode-trace-server) could register itself as a wildcard adopting extension. Do you want it to be prioritized over all other adopting extensions?

AlexDoanBB:
@bhufmann I have rebased and fixed issues with prettier. Thanks!

PatrickTasse (dismissed):
I was able to successfully start and stop the tracecompass-server using this new API. Thanks for the contribution!

I have a question though, if you look at this commit:
https://github.com/PatrickTasse/vscode-trace-server/commit/a772bdc91af95a584995cfaeac12d81bea0effb3
The new API is used blindly (from the IDE, e.g. no code completion or compilation checks) through an 'any' object. Is there a way to have the new API with type-checking available from a VSCode extension, or is this the only way?
Thanks!

AlexDoanBB:
@PatrickTasse Providing type def for the API is possible, either in a new node package or as part of traceviewer-base. Please let us know which way is preferable

vscode-trace-extension/src/extension.ts
```diff
@@ -187,15 +189,23 @@ export function activate(context: vscode.ExtensionContext): ExternalAPI {
...
     traceExtensionWebviewManager.dispose();
 }
 
-async function startTraceServerIfAvailable(): Promise<void> {
+async function startTraceServerIfAvailable(pathToTrace?: string): Promise<void> {
     const extensionId = 'vscode-trace-server';
+    if (await isUp()) {
+        return;
+    }
+    if (pathToTrace) {
```
PatrickTasse, line 204: When using the Open Trace button in the Trace Viewer panel, the command openedTraces.openTraceFolder is invoked and this method is called without a path. The TraceServerManager will not be invoked, and the server will be started using the default tracecompass-community.vscode-trace-server command.

If the trace server manager is to be invoked in this case, the openedTraces.openTraceFolder command handler above should be changed to invoke the user dialog first, then use the chosen path as a parameter to this method.

Alternatively, the Trace Server Manager could be invoked unconditionally with the startServer() parameter changed to being optional, leading to the wildcard contributor (without validator). However I don't know if this will handle your use case because it will start the default generic trace server.

ngondalia: @PatrickTasse  Thank you so much for the review!

You are correct, using "openedTraces.openTraceFolder" will lead to default server being started. I think invoking the user dialog makes more sense than the alternative. WDYT?

 Currently for us, this command is not usable since our traces are a single file trace and not a folder. We do hope to push future contributions to support opening files with that command (Out of the scope of this PR). Out of curiosity, is there any reason why it is currently designed to only open folders?

PatrickTasse (approved):
This implementation is OK to support trace server contributor that is invoked from the "Open with Trace Viewer" context menu in Explorer panel.

However it does not work when using "Open Trace" button because currently the user's selected trace path is not available to provide to the trace server manager. This will need to be improved in a future contribution, by changing to order in that command handler to prompt the user first and then attempt to start the server.

Removing the invocation of start-if-stopped command in favor of using the trace server manager exclusively cannot be done until this improvement is made.

bhufmann (approved):
Thanks @AlexDoanBB  for the contribution. I'm ok with the change and proposed way forward (see https://github.com/eclipse-cdt-cloud/vscode-trace-extension/pull/174#pullrequestreview-1685272995)

vscode-trace-extension/src/extension.ts
ngondalia: Ah ok - Thanks for the explanation!

PatrickTasse:
The improvements mentioned above are implemented here:
https://github.com/eclipse-cdt-cloud/vscode-trace-extension/pull/185

PR #184, opened 2023-10-25
Fix path failure preventing trace from opening

dleclairbb:
When doing some testing on an internal extension, I noticed that we could no longer use the Open With TraceViewer command on Windows due to an invalid path failing (worked on Linux). 

I was able to trace the cause of this it to some code that uses .path instead of .fsPath, which is intended for use with the filesystem and scrubs out improper path values  (which is what caused our error!) I also fixed this in other areas of the code.

This fixes a critical issue which prevents users from opening traces in some cases.

### DIFF
--- a/vscode-trace-extension/src/trace-explorer/trace-tree.ts
+++ b/vscode-trace-extension/src/trace-explorer/trace-tree.ts
@@ -166,8 +166,8 @@ export const fileHandler =
                     return;
                 }
 
-                const uri: string = resolvedTraceURI.path;
-                if (!uri) {
+                const filePath: string = resolvedTraceURI.fsPath;
+                if (!filePath) {
                     traceLogger.addLogMessage(
                         'Cannot open trace: could not retrieve path from URI for trace ' + resolvedTraceURI,
                         fileHandler.name
@@ -175,7 +175,7 @@ export const fileHandler =
                     return;
                 }
 
-                const name = uri.substring(uri.lastIndexOf('/') + 1);
+                const name = path.basename(filePath);
                 const panel = TraceViewerPanel.createOrShow(context.extensionUri, name, undefined);
 
                 progress.report({ message: ProgressMessages.FINDING_TRACES, increment: 10 });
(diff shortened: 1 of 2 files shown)


ngondalia:
@PatrickTasse @bhufmann please review asap. It fixes a breaking change on windows! Thank you :)

vscode-trace-extension/src/trace-explorer/trace-tree.ts
```diff
@@ -311,8 +311,8 @@ const isCtf = async (directory: string): Promise<boolean> => {
 };
 
 function getProgressBarTitle(traceUri: vscode.Uri | undefined): string {
-    if (!traceUri || !traceUri.path) {
+    if (!traceUri || !traceUri.fsPath) {
         return 'undefined';
     }
-    return traceUri.path.substring(traceUri.path.lastIndexOf('/') + 1);
+    return traceUri.fsPath.substring(traceUri.fsPath.lastIndexOf('/') + 1);
```
elaskavaia, line 317: what this suppose to be? basename? Path should use node path module to get file basename

vscode-trace-extension/src/trace-explorer/trace-tree.ts
```diff
@@ -166,7 +166,7 @@ export const fileHandler =
                     return;
                 }
 
-                const uri: string = resolvedTraceURI.path;
+                const uri: string = resolvedTraceURI.fsPath;
```
elaskavaia, line 169: Line 178: const name = uri.substring(uri.lastIndexOf('/') + 1);

That should be basename as well
Also variable should not be called uri, its more like "filePath"

dleclairbb: Fixed in latest commit.

vscode-trace-extension/src/trace-explorer/trace-tree.ts
dleclairbb: Fixed in latest commit.

elaskavaia (approved):

bhufmann (changes requested):
Thanks a lot for the fix!

I tested it on Linux and it looks good. However, the commit message has a reference to what I believe some bug report tool "QVSC-760: fix basename calculation". Could you please remove it?

dleclairbb:
I must have missed that when I was squashing the other changes in - sorry about that! Thank you :)

marcdumais-work (approved):
LGTM, thanks for this contribution! 
Note: I do not have a Windows machine to confirm the bug previous to this contribution, and that it's now fixed. Tested on Linux and the changes seem to work fine still.

bhufmann (approved):
Looks good to me. Thanks for the fix.

PR #185, opened 2023-11-03
Use trace server manager with Open Trace command

### LINKED ISSUE #183: Trace Server Manager cannot be used with Open Trace command
This implementation is OK to support trace server contributor that is invoked from the "Open with Trace Viewer" context menu in Explorer panel.

However it does not work when using "Open Trace" button because currently the user's selected trace path is not available to provide to the trace server manager. This will need to be improved in a future contribution, by changing to order in that command handler to prompt the user first and then attempt to start the server.

Removing the invocation of start-if-stopped command in favor of using the trace server manager exclusively cannot be done until this improvement is made.

_Originally posted by @PatrickTasse in https://github.com/eclipse-cdt-cloud/vscode-trace-extension/pull/174#pullrequestreview-1685272995_

PatrickTasse:
When the command 'openTraceFolder' is invoked from the Open Trace button, change the order so that the user is first prompted to select a trace folder. In order to do this, the openDialog() method is exported and is no longer called from the fileOpenHandler() method.

Use the selected trace to start the trace server using the trace server manager, which will select the proper contributor to start the server, based on the trace path.

Remove the execution of 'start-if-stopped' command from the startTraceServerIfAvailable() method, and make its pathToTrace parameter mandatory.

The fileOpenHandler() method is called last, and its traceUri parameter is now mandatory.

Fixes #183.

### DIFF
--- a/vscode-trace-extension/src/extension.ts
+++ b/vscode-trace-extension/src/extension.ts
@@ -6,6 +6,7 @@ import { TraceExplorerTimeRangeDataProvider } from './trace-explorer/time-range/
 import { TraceExplorerAvailableViewsProvider } from './trace-explorer/available-views/trace-explorer-available-views-webview-provider';
 import { TraceExplorerOpenedTracesViewProvider } from './trace-explorer/opened-traces/trace-explorer-opened-traces-webview-provider';
 import {
+    openDialog,
     fileHandler,
     openOverviewHandler,
     resetZoomHandler,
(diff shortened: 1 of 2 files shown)


PatrickTasse:
https://github.com/eclipse-cdt-cloud/vscode-trace-server/pull/16 should be merged before this one, otherwise the vscode-trace-server will fail to start as it won't be registered as a trace server manager contributor.

bhufmann (changes requested):
Please rebase this change to the latest in master

bhufmann (approved):
It looks good to me. Thanks!

PR #191, opened 2024-01-24
Prepare for first release to OpenVSX

### LINKED ISSUE #69: Deploy to OpenVsx marketplace
Deploy the `vscode-trace-extension` to `OpenVsx` marketplace for users to consume.

marcdumais-work:
# Prepare for first release on openvsx.org

## Changes
 
- Renamed the build workflow to ci-cd and made modifications:
  - added a publish job, that publishes the .vsix extension to openvsx.org,
    triggered only when a release tag (v<something>) is pushed to the repo.
  - merged the build and ui tests jobs. They were redundent, both building
    the project
  - removed yarn package caching: practically it helped very little, cutting
    only ~5% vs the time it takes to install the dependencies without
    ~(25s vs 26s) in the best case (cache hit). With a cache miss (run with 
    modified yarn.lock), we had a 100%+ penalty, if we count the overhead 
    of the cache being saved to GitHub infrastructure ("Post Cache..." job).
- created a "release" workflow that uses "pipe-cd/actions-gh-release" to
  create a release tag and corresponding GitHub release, when a change
  is detected in root file "RELEASE".
    - TL;DR: a new release is triggered by adding the tag value
      reflecting release version, to RELEASE file as part of a PR. The
      action will add tentative release notes to the PR that can be edited
      as needed. Upon merging the PR, a release tag will be created in the
      repo, which will trigger the publish job, that will eventually result
      in the new version of the extension being deployed to openvsx.org
    - it looks like the RELEASE file must exist before its first use in a PR,
      so created one with no active content.
- Added "doc/Publishing.md" to briefly document how to go about creating new
  releases / publish new releases to openvsx.org
- renamed public-facing name (displayName) of the extension to "Trace Viewer
  for VSCode",  to better  reflect its nature
- changed public-facing mentions of "vscode" to the canonical "VSCode"
- changed extension's publisher, to align with the configuration of the CDT
  Cloud OpenVSX publisher user: "eclipse-cdt". See:
  https://www.open-vsx.org/namespace/eclipse-cdt
- other misc updates to the extension's package.json: added/massaged important
  fields
  (https://www.open-vsx.org/namespace/eclipse-cdt).
- kept developer-specific information in repo's root README.md but moved
  generic information, useful to users of the extension, to extension's
  README (which becomes the information page for the extension when published
  or installed).
- Added an icon for the extension. Though probably not the best ever, re-used
  the one from repo "theia-trace-extension"

## Notes

- CI job "Create or prepare GitHub release" failing is normal at this time, because file `./RELEASE` does not yet exist on master. Adding it in this PR will fix this for the future. 
- It's unfortunately not easy to test in advance the publish and release CI jobs. I plan to attempt a first release ASAP to try both, and may find then that some changes are needed. 
- The change in `publisher` to align with `openvsx.org` user for `eclipse-cdt-cloud` may cause momentary issues during the transition, e.g. if a user attempts to install the new version (with the new publisher) when the old one (with original publisher) is already installed. The fact that the extension was never published using the old publisher should mitigate these issues for ordinary end-users.
- companion extension `vscode-trace-server` is not yet published. It would be ideal to also publish it to `openvsx.org` ASAP. It should be possible to do so very similarly as what was done in this PR.

Closes #69

### DIFF
--- a/vscode-trace-extension/README.md
+++ b/vscode-trace-extension/README.md
@@ -1,8 +1,8 @@
 # Trace Viewer for VSCode
 
-This extension adds trace viewing capabilities to VSCode and compatible tools. 
+This extension adds trace viewing capabilities to VSCode and compatible tools. It requires running a trace server, that provides trace information. See [below](#obtain-the-trace-server-eclipse-trace-compass) for more details.
 
-For information about building this extension from sources, debugging, etc, please see [Developing.md](https://github.com/eclipse-cdt-cloud/vscode-trace-extension/tree/master/README.md)
+For information about building this extension from sources, debugging, etc, please see [here (README.md)](https://github.com/eclipse-cdt-cloud/vscode-trace-extension/tree/master/README.md)
 
 ## Using the extension
 
@@ -21,11 +21,14 @@ The `Views` tab shows all the available views for the selected trace. Click on a
 [open-output]: https://raw.githubusercontent.com/eclipse-cdt-cloud/vscode-trace-extension/master/doc/images/vscode-trace-extension-001.png
 [open-trace]: https://raw.githubusercontent.com/eclipse-cdt-cloud/vscode-trace-extension/master/doc/images/vscode-open-with-trace-viewer-001.png
 
+<a name="trace-server"></a>
+
 ## Obtain the Trace Server (Eclipse Trace Compass)
 
-In order to open and view traces, you need a trace server running on the same machine as the trace extension. You can use the Eclipse Trace Compass server:
+In order to open and view traces, you need a trace server, that supports TSP (Trace Server Protocol), running on the same machine as this extension. You can use the Eclipse Trace Compass server:
 
 Download the Latest "incubator" build:
+
 - [Linux x86_64](https://download.eclipse.org/tracecompass.incubator/trace-server/rcp/trace-compass-server-latest-linux.gtk.x86_64.tar.gz)
 - [other Operating Systems / Architectures](https://download.eclipse.org/tracecompass.incubator/trace-server/rcp/)
 


.github/workflows/release.yml
```diff
@@ -0,0 +1,29 @@
...
+        branches:    
+          - master
+        paths:
+          - 'RELEASE'
+    pull_request:
+        types: [opened, synchronize]
+        branches:
+          - master
+        paths:
+          - 'RELEASE'
```
bhufmann, line 14: Will that run the release already when a pull request is created?

bhufmann (changes requested):

.github/workflows/release.yml
marcdumais-work: I think it will trigger the action but when it checks the RELEASE file it should immediately exit since there is no "tag" entry in the file yet.

vscode-trace-extension/README.md
```diff
@@ -0,0 +1,38 @@
+# Trace Viewer for VSCode
+
+This extension adds trace viewing capabilities to VSCode and compatible tools. 
+
+For information about building this extension from sources, debugging, etc, please see [Developing.md](https://github.com/eclipse-cdt-cloud/vscode-trace-extension/tree/master/README.md)
+
+## Using the extension
```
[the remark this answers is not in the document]
marcdumais-work: I just noticed above a left-over, from a previous version of the PR: the mention of "Developing.md". I will correct this.

.github/workflows/release.yml
marcdumais-work: Oops - I answered the wrong question, sorry. 

When a new PR is created after this is merged, the action will add the automatically created release notes to the PR, leaving the opportunity to make changes if desired (e.g. by using more detailed options in file `RELEASE`). The GitHub release and release tags are only created upon a PR being merged.

marcdumais-work: More info about how this action works: 
https://github.com/pipe-cd/actions-gh-release#usage

vscode-trace-extension/README.md
```diff
@@ -0,0 +1,38 @@
+# Trace Viewer for VSCode
+
+This extension adds trace viewing capabilities to VSCode and compatible tools. 
+
```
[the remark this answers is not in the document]
marcdumais-work: Done - see latest pushed commit. Let me know if the added sentence is lacking.

vscode-trace-extension/README.md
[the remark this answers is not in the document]
marcdumais-work: I have corrected the leftover above in the latest commit.

marcdumais-work:
I just pushed a rebase on master

PR #192, opened 2024-01-25
Fix 'yarn watch' Script

williamsyang-work:
## Fix 'Yarn Watch' Script

Yarn watch script in root now executes 'yarn run watch' in each subdirectory and all outputs are streamed to a single terminal.
Tested to ensure that the changes in each directory are shown in Extension Development Host (webviews, extension, common) with the script.
README updated to reflect changes.

Developer QOL improvement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,9 +60,7 @@ ln -s <vscode-trace-extension root>/vscode-trace-extension-x.x.x.vsix ./
 
 ## Developing the extension
 
-When having to modify the code of the extension (in the `vscode-trace-extension` folder), one can simply run the `yarn` command. It is also possible to watch for changes to have no manual steps to do before re-running the extension: `yarn watch` or `ctrl-shift-b` and select the task `npm: watch - vscode-trace-extension`.
-
-For changes in the webview part (in the `vscode-trace-webviews` folder), you can run the `yarn` command, simply re-opening a trace should show the changes. It is also possible to watch for changes with `yarn watch` or `ctrl-shift-b` and selecting the task `npm: watch - vscode-trace-webviews`.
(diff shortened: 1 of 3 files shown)


marcdumais-work:
Thanks for this PR! Good idea having one script at the root of the repo to watch the various components. 

I think it might be nice to have two such scripts: one to watch compilation of all component (tsc -w) and one to watch the bundling (webpack --watch) for relevant components. `tsc -w` has a quite terse output, so the output of watching compilation would more easily fit on one page of the terminal. But it would not catch webpack-specific issues.

However, It think this PR is an improvement in itself and having a second watch script could be done separately at a later time (not necessarily by you unless you want-to). WDYT?

williamsyang-work:
> Thanks for this PR! Good idea having one script at the root of the repo to watch the various components.
> 
> I think it might be nice to have two such scripts: one to watch compilation of all component (tsc -w) and one to watch the bundling (webpack --watch) for relevant components. `tsc -w` has a quite terse output, so the output of watching compilation would more easily fit on one page of the terminal. But it would not catch webpack-specific issues.
> 
> That being said, It think this PR is an improvement in itself and having a second watch script could be done separately at a later time (not necessarily by you unless you want-to). WDYT?

I agree, the webpack output is cluttered.

`tsc -w` by itself will not reflect changes for 'extension' and 'webview' in the extension developer.  They gotta be webpack'd.
This watch script calls `tsc -w` in 'common' and webpack --watch in 'extension' and 'webview'.
 
We could run all `webpack --watch` commands under root `watch:bundle`  and all `tsc -w` commands under plain `watch` and tell contributors to run both when developing.

williamsyang-work:
I think we can leave it as-is for now and change it in the future if it comes up!  What do you think?

marcdumais-work (approved):
LGTM, thanks @williamsyang-work !

I do not think this change will be controversial, so I will go ahead and merge it.

PR #193, opened 2024-01-29
Update eslint dependency to >= 8.43.0

marcdumais-work:
Earlier versions contain content that's licensed under an unapproved license. I adjusted the version we request in a few places.

An older version was still being pulled from one of our dependencies, `rewire`. Upon searching the sources, I could not find where this dependency was being used and so I removed it.

Also updated some `eslint`-related packages that were quite old.


bhufmann (approved):
LGTM. Thanks for update!

PR #194, opened 2024-01-30
Trigger first release to open-vsx.org

marcdumais-work:
This is the first try, so we may need to adjust some things to make it work in the end.

The idea is to update file `RELEASE`, adding a new release tag in there. Then we expect that tentative, automatically-generated release notes, will be added to this PR, for our review. Once we merge the PR, a GitHub release v0.2.0 should be created automatically in the repo, with attached release notes. The tag's creation should then trigger the job that published the extension to open-vsx.org.

Note: I picked v0.2.0 to avoid conflicting with the "vscode trace extension", that's already released on the Visual Studio Marketplace (which is currently at v0.1.44). I propose we go with a v1.0.0 version after discussions with other contributors, and after our release/publish infrastructure has been demonstrated to work well.

See [below](https://github.com/eclipse-cdt-cloud/vscode-trace-extension/pull/194#issuecomment-1920084882) for the automatically generated release notes we are currently obtaining. I played-around with the [options](https://github.com/pipe-cd/actions-gh-release/tree/v2.3.4/#usage) quite a bit and something about them is flaky - it could be due to the slightly strange initial run we have, going-back a long while in history, for this first run. In any case, the generation does not need to be perfect -any committer can edit them after the fact, from the GitHub release page. 

I think it might be a bit controversial to have both commits and PRs documented in there - I personally like the idea of having one page where we can quickly find both, but I suggest we adjust the options after merging this PR, at the next release - I hope it will work better then.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,7 +22,7 @@ jobs:
             - uses: actions/checkout@v3
               with:
                 fetch-depth: 0
-            - uses: pipe-cd/actions-gh-release@v2.3.4
+            - uses: pipe-cd/actions-gh-release@v2.6.0
               with:
                 release_file: 'RELEASE'
                 token: ${{ secrets.GITHUB_TOKEN }}
--- a/RELEASE
+++ b/RELEASE
@@ -1,1 +1,8 @@
-tag: v0.0.0
+tag: v0.2.0
+
+commitInclude:
+  parentOfMergeCommit: true
+
+releaseNoteGenerator:
+  showAbbrevHash: true
+  showCommitter: false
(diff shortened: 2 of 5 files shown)


marcdumais-work:
Note: We will need to change RELEASE to add a bogus, previous to `v0.2.0` release tag file before merging this PR, for `actions-gh-release` to work.

marcdumais-work:
> Note: We will need to change RELEASE to add a bogus, previous to `v0.2.0` release tag file before merging this PR, for `actions-gh-release` to work.

done in #197

MatthewKhouzam (approved):
lgtm

marcdumais-work:
Thanks for the quick review!

PR #198, opened 2024-01-31
[CI] Use 'list' Playwright reporter

marcdumais-work:
When running the tests locally, the current 'html' reporter is fine. But I've bveen confused with it in CI logs where the was a failure, presumably a silent retry and finally a report of the tests passing.

I think in CI we would have more clarity with an alternate reporter. Let's try 'list'. If it does not work well we can fall-back on the simple 'dot'.


marcdumais-work:
We got lucky: we got the case I was trying to improve in the latest CI run (a failure followed by a report of overall success):
https://github.com/eclipse-cdt-cloud/vscode-trace-extension/actions/runs/7742470139/job/21111681510?pr=198#step:14:24

With the new reporter, we can see just above a report of the test running:
https://github.com/eclipse-cdt-cloud/vscode-trace-extension/actions/runs/7742470139/job/21111681510?pr=198#step:14:15

marcdumais-work:
I think maybe the (default?) timeout of 5000ms [might be a little short](https://github.com/eclipse-cdt-cloud/vscode-trace-extension/actions/runs/7742470139/job/21111681510?pr=198#step:14:30)

PR #200, opened 2024-02-05
Use GitHub token from a committer in gh-release job

marcdumais-work:
It turns-out that using the GitHub token, that's automatically generated for each workflow run, does not allow for further workflow(s) being triggered. We were relying on that feature, for the "publish" job to trigger, after action "actions-gh-release" creates a git tag in the repo.
    
Instead, for that to work, we need to use an alternate token, from a committer. I will set such a token in the repo as a secret, under name "GH_COMMITTER_TOKEN"

Also a couple of minor upadates to the `ci-cd` workslow:
- use 'release' as workflow trigger to start publish job
- check before running job that the workflow was triggered by a release tag 'reference". These map to, e.g.: 

  github.event_name:  release
  github.ref:  refs/tags/v0.2.2

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -5,8 +5,9 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ master ]
-  # branch or tag created  
-  create:
+  release:
+    types:
+      - published
 
 jobs:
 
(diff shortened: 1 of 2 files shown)


bhufmann (approved):
LGTM. Thanks!

PR #203, opened 2024-02-13
[CI] Infrastructure to enable publish to Marketplace 

marcdumais-work:
Add a job to publish the extension to the Visual Studio Marketplace, using the PAT setup for us by webmaster at org level.

This PR will not, by itself, trigger a publish to the Marketplace. Rather, the new job will trigger along with the existing job that publishes to the public open-vsx registry, when an eventual PR is merged, that modifies the `RELEASE` root file, adding a new release tag in there.

Note: Since the [Trace Server extension](https://github.com/eclipse-cdt-cloud/vscode-trace-server), depends on this extension here, we should be careful to publish this (Viewer) extension first to the Marketplace.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -84,8 +84,9 @@ jobs:
     - name: Run format check
       run: yarn format:check
 
-  publish:
-    name: Publish extension to openvsx.org
+  publish-oxsv:
+    # https://open-vsx.org/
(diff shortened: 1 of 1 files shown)


marcdumais-work:
Looks like the publish token is setup correctly:
image

bhufmann (approved):
LGTM. Thanks!

marcdumais-work:
Thanks for the review!

PR #196, opened 2024-01-30
Use asExternalUri For Server Url

williamsyang-work:
Refactors TspClientProvider to use VSCode.asExternalUri method to generate a dynamic
uri.  This will automatically detect if we are using remote-ssh, initialize port forwarding
to the server's port, and change the provided url to a web-worker that is managing
port-forwarding to the VSCode backend.

Removes the TraceServerUrlManager class.  The functionality to generate new TspClients
when the Uri is changed is moved into this repo's TspClientProvider implementation.

Removes the tspClient.ts file.  This functionality did not work with the async nature of
asExternalUri, and was not designed to handle Uri changes.  Most functionality was already
implemented in TspClientProviderImpl.ts.

Creates VSCodeBackendTspClientProvider class.  This is a child class of TspClientProviderImpl
with expanded functionality that should only be used on the extension's backend.

Signed-off-by: Will Yang <william.yang@ericsson.com>

PR #204, opened 2024-02-14
Proper Handling of the URL Configuration Change

ngondalia:
This commit provides proper handling of url configuration change. This fixes the issue of when the user changes the URL of the server, the user would need to reopen vscode so that all of the manangers can be initialized with the proper tspClientURL. With this commit, all tspClient dependent components are notified of a URL change.

Signed-off-by: Neel Gondalia ngondalia@blackberry.com

### DIFF
--- a/vscode-trace-extension/src/trace-explorer/available-views/trace-explorer-available-views-webview-provider.ts
+++ b/vscode-trace-extension/src/trace-explorer/available-views/trace-explorer-available-views-webview-provider.ts
@@ -33,6 +33,7 @@ export class TraceExplorerAvailableViewsProvider implements vscode.WebviewViewPr
     public updateTraceServerUrl(newUrl: string): void {
         if (this._view) {
             this._view.webview.postMessage({ command: VSCODE_MESSAGES.TRACE_SERVER_URL_CHANGED, data: newUrl });
+            this._view.webview.html = this._getHtmlForWebview(this._view.webview);
         }
     }
 
--- a/vscode-trace-extension/src/trace-explorer/opened-traces/trace-explorer-opened-traces-webview-provider.ts
+++ b/vscode-trace-extension/src/trace-explorer/opened-traces/trace-explorer-opened-traces-webview-provider.ts
@@ -56,6 +56,7 @@ export class TraceExplorerOpenedTracesViewProvider implements vscode.WebviewView
     public updateTraceServerUrl(newUrl: string): void {
         if (this._view) {
             this._view.webview.postMessage({ command: VSCODE_MESSAGES.TRACE_SERVER_URL_CHANGED, data: newUrl });
+            this._view.webview.html = this._getHtmlForWebview(this._view.webview);
         }
     }
 
--- a/vscode-trace-extension/src/trace-explorer/properties/trace-explorer-properties-view-webview-provider.ts
+++ b/vscode-trace-extension/src/trace-explorer/properties/trace-explorer-properties-view-webview-provider.ts
@@ -4,6 +4,7 @@
  * Licensed under the MIT license. See LICENSE file in the project root for details.
  ***************************************************************************************/
 import * as vscode from 'vscode';
+import { VSCODE_MESSAGES } from 'vscode-trace-common/lib/messages/vscode-message-manager';
 import { traceExtensionWebviewManager } from 'vscode-trace-extension/src/extension';
 import { getTraceServerUrl } from 'vscode-trace-extension/src/utils/tspClient';
 
@@ -33,6 +34,13 @@ export class TraceExplorerItemPropertiesProvider implements vscode.WebviewViewPr
     postMessagetoWebview(_command: string, _data: unknown): void {
         if (this._view && _command && _data) {
             this._view.webview.postMessage({ command: _command, data: _data });
+        }
+    }
+
+    public updateTraceServerUrl(newUrl: string): void {
+        if (this._view) {
+            this._view.webview.postMessage({ command: VSCODE_MESSAGES.TRACE_SERVER_URL_CHANGED, data: newUrl });
+            this._view.webview.html = this._getHtmlForWebview(this._view.webview);
         }
     }
 
--- a/vscode-trace-extension/src/trace-explorer/time-range/trace-explorer-time-range-data-webview-provider.ts
+++ b/vscode-trace-extension/src/trace-explorer/time-range/trace-explorer-time-range-data-webview-provider.ts
@@ -82,6 +82,13 @@ export class TraceExplorerTimeRangeDataProvider implements vscode.WebviewViewPro
         );
     }
 
+    public updateTraceServerUrl(newUrl: string): void {
+        if (this._view) {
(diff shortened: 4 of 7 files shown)


williamsyang-work:
Hey Neel,

I currently have a PR under review that addresses this exact problem as well.  https://github.com/eclipse-cdt-cloud/vscode-trace-extension/pull/196.

Maybe you can take a look and see if this works for you?

bhufmann (commented):
I was testing this change and the URL change is propagated correctly. However, I noticed that when change the URL, the webviews don't populate and I see the error message below in the console log. In my test I changed the URL from `localhost` to `10.0.0.122`. All the updates are done but, the html of the webview is not resolve again and it has the old URL for the `connect-src`. For example, the method `resolveWebviewView()` which calls `_getHtmlForWebview` of `TraceExplorerOpenedTracesViewProvider` is only called once and hence the `connect-src` is not updated. I'm not sure how to fix that. Please note it's a general problem and not related to this change. Any ideas how to fix it?

image

ngondalia:
> I was testing this change and the URL change is propagated correctly. However, I noticed that when change the URL, the webviews don't populate and I see the error message below in the console log. In my test I changed the URL from `localhost` to `10.0.0.122`. All the updates are done but, the html of the webview is not resolve again and it has the old URL for the `connect-src`. For example, the method `resolveWebviewView()` which calls `_getHtmlForWebview` of `TraceExplorerOpenedTracesViewProvider` is only called once and hence the `connect-src` is not updated. I'm not sure how to fix that. Please note it's a general problem and not related to this change. Any ideas how to fix it?
> 
> image

I pushed a patch for it! Can you please re-test it and let me know? 

From the docs of webview.html property:
"
HTML contents of the webview.

This should be a complete, valid html document. Changing this property causes the webview to be reloaded.
"

Based on this, you can just assign the updated html content to the webview and it will get reloaded.

bhufmann:
> > I was testing this change and the URL change is propagated correctly. However, I noticed that when change the URL, the webviews don't populate and I see the error message below in the console log. In my test I changed the URL from `localhost` to `10.0.0.122`. All the updates are done but, the html of the webview is not resolve again and it has the old URL for the `connect-src`. For example, the method `resolveWebviewView()` which calls `_getHtmlForWebview` of `TraceExplorerOpenedTracesViewProvider` is only called once and hence the `connect-src` is not updated. I'm not sure how to fix that. Please note it's a general problem and not related to this change. Any ideas how to fix it?
> > image
> 
> I pushed a patch for it! Can you please re-test it and let me know?
> 
> From the docs of webview.html property: " HTML contents of the webview. This should be a complete, valid html document. Changing this property causes the webview to be reloaded. "
> 
> Based on this, you can just assign the updated html content to the webview and it will get reloaded.

Yes, this did the trick. I tested it and works. Thanks.

I'm going to do a review now. Once this PR merged, @williamsyang-work will base his work on it.

vscode-trace-extension/src/trace-explorer/available-views/trace-explorer-available-views-webview-provider.ts
```diff
@@ -36,6 +36,10 @@ export class TraceExplorerAvailableViewsProvider implements vscode.WebviewViewPr
         }
     }
 
+    public rerender() {
```
bhufmann, line 39: I think you can call the lines below in method `updateTraceServerUrl()`. With this you can avoid adding this public API.

vscode-trace-extension/src/trace-explorer/opened-traces/trace-explorer-opened-traces-webview-provider.ts
```diff
@@ -59,6 +59,10 @@ export class TraceExplorerOpenedTracesViewProvider implements vscode.WebviewView
         }
     }
 
+    public rerender() {
```
bhufmann, line 62: I think you can call the lines below in method `updateTraceServerUrl()`. With this you can avoid adding this public API.

vscode-trace-extension/src/trace-explorer/properties/trace-explorer-properties-view-webview-provider.ts
```diff
@@ -12,6 +12,10 @@ export class TraceExplorerItemPropertiesProvider implements vscode.WebviewViewPr
     private _view?: vscode.WebviewView;
     constructor(private readonly _extensionUri: vscode.Uri) {}
 
+    public rerender() {
```
bhufmann, line 15: I think you can call the lines below in method `updateTraceServerUrl()`. With this you can avoid adding this public API.

vscode-trace-extension/src/trace-explorer/time-range/trace-explorer-time-range-data-webview-provider.ts
```diff
@@ -20,6 +20,10 @@ export class TraceExplorerTimeRangeDataProvider implements vscode.WebviewViewPro
         this._experimentDataMap = new TimeRangeDataMap();
     }
 
+    public rerender() {
```
bhufmann, line 23: I think you can call the lines below in method `updateTraceServerUrl()`. With this you can avoid adding this public API.

vscode-trace-extension/src/trace-viewer-panel/trace-viewer-webview-panel.ts
```diff
@@ -70,6 +70,14 @@ export class TraceViewerPanel {
         return openedPanel;
     }
 
+    public static rerenderPanels() {
+        Object.values(TraceViewerPanel.activePanels).forEach(panel => {
```
bhufmann, line 74: I think you can call the lines below in method `updateTraceServerUrl()`. With this you can avoid adding this public API.

bhufmann (changes requested):

bhufmann (approved):
Looks good to me. Thanks a lot for this contribution.

ngondalia:
> Looks good to me. Thanks a lot for this contribution. Since you're committer, you can submit it. Please use `Rebase and Merge`.

I don't have that option. It seems like i still don't have write access.

williamsyang-work:
> > I was testing this change and the URL change is propagated correctly. However, I noticed that when change the URL, the webviews don't populate and I see the error message below in the console log. In my test I changed the URL from `localhost` to `10.0.0.122`. All the updates are done but, the html of the webview is not resolve again and it has the old URL for the `connect-src`. For example, the method `resolveWebviewView()` which calls `_getHtmlForWebview` of `TraceExplorerOpenedTracesViewProvider` is only called once and hence the `connect-src` is not updated. I'm not sure how to fix that. Please note it's a general problem and not related to this change. Any ideas how to fix it?
> > image
> 
> I pushed a patch for it! Can you please re-test it and let me know?
> 
> From the docs of webview.html property: " HTML contents of the webview. This should be a complete, valid html document. Changing this property causes the webview to be reloaded. "
> 
> Based on this, you can just assign the updated html content to the webview and it will get reloaded.

Awesome find.  This bug has been on my radar for a while, I appreciate the fix!

hriday-panchasara (approved):
LGTM, thanks!

PR #213, opened 2024-02-29
[TENTATIVE - DO NOT MERGE][Breaking] Rename extension to: eclipse-cdt…

marcdumais-work:
….vscode-trace-viewer

Note: this is tentative for now - we will attempt to publish to the Marketplace without changing the name at first.

From 'vscode-trace-extension'. This affects the extension's name, used to publish it. The internal name, e.g. used in the repo's folder name keep the original name.

This name change is necessary so we can publish the extension to the Visual Studio Marketplace, which now requires that extensions published there have a unique name, irrespective of publisher/namespace.

Note: this is a breaking change since other extensions that use the API offered by this one will need to get a reference, using the vscode extensions API, using the new name to look it up.

For example, companion extension "eclipse-cdt.vscode-trace-server", which does that, will need to use the new name.

 Follow-up: will need to adjust any reference from companion extension vscode-trace-viewer, that uses the API. See tentative PR:
https://github.com/eclipse-cdt-cloud/vscode-trace-server/pull/31

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# VSCode Trace Extension
+# VSCode Trace Viewer Extension
 
 This document contains information that may be useful for developers that want to build, modify, enhance and/or debug this extension. If you only intend to consume the extension, it might be easier to get it from the [public OpenVSX registry](https://www.open-vsx.org/extension/eclipse-cdt/vscode-trace-extension),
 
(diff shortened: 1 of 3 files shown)


vscode-trace-extension/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "vscode-trace-extension",
+  "name": "vscode-trace-viewer",
```
marcdumais-work, line 2: alternative: `vscode-trace-viewer-extension`

README.md
```diff
@@ -232,7 +232,7 @@ onWebviewPanelCreated(listener: (data: vscode.WebviewPanel) => void): void
 
 ```javascript
 //The following retrieves the API object from the vscode-trace-extension
```
marcdumais-work, line 234: forgot to update in the comment

marcdumais-work:
Un-publishing the extension from the Marketplace worked, so no need to proceed with this PR.

PR #215, opened 2024-03-08
[license] Get rid of version of 3PP version that's deprecated and fails the license check

marcdumais-work:
The following 3PP dependency version, used in this repo and that currently fails the license check, has been deprecated:

`@fortawesome/fontawesome-common-types@0.3.0`

Eclipse Foundation ticket about its license:

https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/12947

In theia-trace-extension, we had previously constrained the version range of the related dependency that pulls it, forcing a non-deprecated version to be used:

https://github.com/eclipse-cdt-cloud/theia-trace-extension/commit/e44236a292b3125e95454944c1ed8199e0692f52

I think we should do the same here.

### DIFF
--- a/vscode-trace-extension/package.json
+++ b/vscode-trace-extension/package.json
@@ -259,7 +259,7 @@
     ]
   },
   "dependencies": {
-    "@fortawesome/fontawesome-svg-core": "^1.2.17",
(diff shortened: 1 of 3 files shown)


MatthewKhouzam (approved):
Great catch! it was a legit failure!

PR #216, opened 2024-03-08
Improvements for accurate display of 'Welcome View' 

williamsyang-work:
Adds `traceViewer.serverOn` VSCode context.

Replaces the use of `ReactPlaceholderWidget` with a welcome view.

Improves tracking of `trace-explorer.noExperiments` VSCode Context.

image

### DIFF
--- a/vscode-trace-extension/package.json
+++ b/vscode-trace-extension/package.json
@@ -90,6 +90,11 @@
         "icon": "$(new-folder)"
       },
       {
+        "command": "trace-explorer.refreshContext",
+        "title": "Refresh Trace Explorer",
+        "icon": "$(refresh)"
+      },
+      {
         "command": "traceViewer.shortcuts",
         "title": "Trace Viewer Keyboard and Mouse Shortcuts",
         "icon": "$(info)"
@@ -175,6 +180,11 @@
         {
           "command": "openedTraces.openTraceFolder",
           "when": "view == traceExplorer.openedTracesView && !trace-explorer.noExperiments",
+          "group": "navigation"
+        },
+        {
+          "command": "trace-explorer.refreshContext",
+          "when": "view == welcome || view == traceExplorer.openedTracesView",
           "group": "navigation"
         }
       ],
--- a/vscode-trace-extension/package.json
+++ b/vscode-trace-extension/package.json
@@ -125,42 +125,49 @@
     "views": {
       "trace-explorer": [
         {
+          "type": "tree",
+          "name": "Welcome",
+          "id": "welcome",
+          "when": "trace-explorer.noExperiments || !traceViewer.serverUp"
+        },
+        {
           "type": "webview",
           "id": "traceExplorer.openedTracesView",
-          "name": "Opened Traces"
+          "name": "Opened Traces",
+          "when": "!trace-explorer.noExperiments && traceViewer.serverUp"
         },
         {
           "type": "webview",
           "id": "traceExplorer.availableViews",
           "name": "Views",
-          "when": "!trace-explorer.noExperiments"
+          "when": "!trace-explorer.noExperiments && traceViewer.serverUp"
         },
         {
           "type": "webview",
           "id": "traceExplorer.timeRangeDataView",
           "name": "Time Range Data",
-          "when": "!trace-explorer.noExperiments"
+          "when": "!trace-explorer.noExperiments && traceViewer.serverUp"
         },
         {
           "type": "webview",
           "id": "traceExplorer.itemPropertiesView",
           "name": "Item Properties",
-          "when": "!trace-explorer.noExperiments"
+          "when": "!trace-explorer.noExperiments && traceViewer.serverUp"
         }
       ]
     },
     "viewsWelcome": [
       {
-        "view": "traces",
-        "contents": "No traces found [learn more](https://www.eclipse.org/tracecompass/).\n[Add traces](command:nodeDependencies.addEntry)"
+        "view": "welcome",
+        "name": "Open Trace",
+        "contents": "There are currently no opened traces. \n[Open Trace](command:openedTraces.openTraceFolder)",
+        "when": "trace-explorer.noExperiments || !traceViewer.serverUp"
       },
       {
-        "view": "views",
-        "contents": "No trace selected [learn more](https://www.eclipse.org/tracecompass/)"
-      },
-      {
-        "view": "item properties",
-        "contents": "No properties to show"
+        "view": "welcome",
+        "name": "Documentation",
+        "contents": "To learn more about how to use the Trace Viewer [read our documentation](https://github.com/eclipse-cdt-cloud/vscode-trace-extension/blob/master/vscode-trace-extension/README.md).",
+        "when": "trace-explorer.noExperiments || !traceViewer.serverUp"
       }
     ],
     "menus": {
--- a/vscode-trace-extension/src/test/extension-test.spec.ts
+++ b/vscode-trace-extension/src/test/extension-test.spec.ts
@@ -16,7 +16,7 @@ test('Open Trace from Explorer', async ({ page }) => {
 
 test('Open Trace from Trace Viewer', async ({ page }) => {
     await page.getByRole('tab', { name: 'Trace Viewer' }).locator('a').click();
-    await page.getByRole('button', { name: 'Opened Traces Section' }).hover();
+    await page.getByRole('button', { name: 'Open Trace' }).hover();
     await page.getByRole('button', { name: 'Open Trace' }).click();
     await page.getByRole('option', { name: '202-bug-hunt' }).locator('a').click();
     await page.getByRole('option', { name: 'cat-kernel' }).locator('a').click();
--- a/vscode-trace-extension/src/utils/backend-tsp-client-provider.ts
+++ b/vscode-trace-extension/src/utils/backend-tsp-client-provider.ts
@@ -36,12 +36,15 @@ export const addTspClientChangeListener = (listenerFunction: (tspClient: TspClie
 
 /**
  * Get the status of the server.
+ * Updates the VSCode Context for `traceViewer.serverUp`
  * @returns server status as boolean
  */
-export async function isUp(): Promise<boolean> {
+export async function isTraceServerUp(): Promise<boolean> {
     const health = await getTspClient().checkHealth();
     const status = health.getModel()?.status;
-    return health.isOk() && status === 'UP';
+    const serverIsUp = health.isOk() && status === 'UP';
+    vscode.commands.executeCommand('setContext', 'traceViewer.serverUp', serverIsUp);
+    return serverIsUp;
 }
 
 function getUriRootFromUserSettings(): string {
--- a/vscode-trace-extension/src/utils/backend-tsp-client-provider.ts
+++ b/vscode-trace-extension/src/utils/backend-tsp-client-provider.ts
@@ -47,6 +47,19 @@ export async function isTraceServerUp(): Promise<boolean> {
     return serverIsUp;
 }
 
+/**
+ * Checks for opened traces and updates the vscode context `trace-explorer.noExperiments`
+ */
+export async function updateNoExperimentsContext(): Promise<void> {
+    const response = await getTspClient().fetchExperiments();
+    if (!response.isOk()) {
+        return;
+    }
+    const noExperiments = !response.getModel()?.length;
+    vscode.commands.executeCommand('setContext', 'trace-explorer.noExperiments', noExperiments);
+    return;
+}
+
 function getUriRootFromUserSettings(): string {
     const tsConfig = vscode.workspace.getConfiguration('trace-compass.traceserver');
     const traceServerUrl: string = tsConfig.get<string>('url') || 'http://localhost:8080';
--- a/vscode-trace-extension/src/extension.ts
+++ b/vscode-trace-extension/src/extension.ts
@@ -15,7 +15,12 @@ import {
     keyboardShortcutsHandler
 } from './trace-explorer/trace-tree';
 import { TraceServerConnectionStatusService } from './utils/trace-server-status';
-import { getTspClientUrl, updateTspClientUrl, isTraceServerUp } from './utils/backend-tsp-client-provider';
+import {
+    getTspClientUrl,
+    updateTspClientUrl,
+    isTraceServerUp,
+    updateNoExperimentsContext
+} from './utils/backend-tsp-client-provider';
 import { TraceExtensionLogger } from './utils/trace-extension-logger';
 import { ExternalAPI, traceExtensionAPI } from './external-api/external-api';
 import { TraceExtensionWebviewManager } from './utils/trace-extension-webview-manager';
@@ -190,9 +195,20 @@ export function activate(context: vscode.ExtensionContext): ExternalAPI {
         })
     );
 
+    context.subscriptions.push(
+        vscode.commands.registerCommand('trace-explorer.refreshContext', async () => {
+            const isUp = await isTraceServerUp();
+            vscode.commands.executeCommand('setContext', 'traceViewer.serverUp', isUp);
+            if (isUp) {
+                await updateNoExperimentsContext();
+            }
+        })
+    );
+
     vscode.commands.executeCommand('setContext', 'traceViewer.markerSetsPresent', false);
     vscode.commands.executeCommand('setContext', 'traceViewer.markerCategoriesPresent', false);
     vscode.commands.executeCommand('setContext', 'trace-explorer.noExperiments', true);
+    vscode.commands.executeCommand('trace-explorer.refreshContext');
     serverStatusService.checkAndUpdateServerStatus();
     return traceExtensionAPI;
 }
--- a/vscode-trace-extension/src/extension.ts
+++ b/vscode-trace-extension/src/extension.ts
@@ -74,6 +74,7 @@ export function activate(context: vscode.ExtensionContext): ExternalAPI {
             await startTraceServerIfAvailable(file.fsPath);
             if (await isTraceServerUp()) {
                 fileOpenHandler(context, file);
+                vscode.commands.executeCommand('setContext', 'trace-explorer.noExperiments', false);
             }
         })
     );
@@ -163,6 +164,7 @@ export function activate(context: vscode.ExtensionContext): ExternalAPI {
             await startTraceServerIfAvailable(traceUri.fsPath);
             if (await isTraceServerUp()) {
                 fileOpenHandler(context, traceUri);
+                vscode.commands.executeCommand('setContext', 'trace-explorer.noExperiments', false);
             }
         })
     );
@@ -190,6 +192,7 @@ export function activate(context: vscode.ExtensionContext): ExternalAPI {
 
     vscode.commands.executeCommand('setContext', 'traceViewer.markerSetsPresent', false);
     vscode.commands.executeCommand('setContext', 'traceViewer.markerCategoriesPresent', false);
+    vscode.commands.executeCommand('setContext', 'trace-explorer.noExperiments', true);
     serverStatusService.checkAndUpdateServerStatus();
     return traceExtensionAPI;
 }
--- a/vscode-trace-extension/src/extension.ts
+++ b/vscode-trace-extension/src/extension.ts
@@ -15,7 +15,7 @@ import {
     keyboardShortcutsHandler
 } from './trace-explorer/trace-tree';
 import { TraceServerConnectionStatusService } from './utils/trace-server-status';
-import { getTspClientUrl, updateTspClientUrl, isUp } from './utils/backend-tsp-client-provider';
+import { getTspClientUrl, updateTspClientUrl, isTraceServerUp } from './utils/backend-tsp-client-provider';
 import { TraceExtensionLogger } from './utils/trace-extension-logger';
 import { ExternalAPI, traceExtensionAPI } from './external-api/external-api';
 import { TraceExtensionWebviewManager } from './utils/trace-extension-webview-manager';
@@ -72,7 +72,7 @@ export function activate(context: vscode.ExtensionContext): ExternalAPI {
     context.subscriptions.push(
         vscode.commands.registerCommand('traces.openTraceFile', async (file: vscode.Uri) => {
             await startTraceServerIfAvailable(file.fsPath);
-            if (await isUp()) {
+            if (await isTraceServerUp()) {
                 fileOpenHandler(context, file);
             }
         })
@@ -161,7 +161,7 @@ export function activate(context: vscode.ExtensionContext): ExternalAPI {
                 return;
             }
             await startTraceServerIfAvailable(traceUri.fsPath);
-            if (await isUp()) {
+            if (await isTraceServerUp()) {
                 fileOpenHandler(context, traceUri);
             }
         })
@@ -175,7 +175,7 @@ export function activate(context: vscode.ExtensionContext): ExternalAPI {
 
     context.subscriptions.push(
         vscode.commands.registerCommand('serverStatus.started', () => {
-            serverStatusService.render(true);
+            serverStatusService.checkAndUpdateServerStatus();
             if (tracesProvider) {
                 tracesProvider.postMessagetoWebview(VSCODE_MESSAGES.TRACE_SERVER_STARTED, undefined);
             }
@@ -184,12 +184,13 @@ export function activate(context: vscode.ExtensionContext): ExternalAPI {
 
     context.subscriptions.push(
         vscode.commands.registerCommand('serverStatus.stopped', () => {
-            serverStatusService.render(false);
+            serverStatusService.checkAndUpdateServerStatus();
         })
     );
 
     vscode.commands.executeCommand('setContext', 'traceViewer.markerSetsPresent', false);
     vscode.commands.executeCommand('setContext', 'traceViewer.markerCategoriesPresent', false);
+    serverStatusService.checkAndUpdateServerStatus();
     return traceExtensionAPI;
 }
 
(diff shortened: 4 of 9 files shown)


vscode-trace-extension/src/test/extension-test.spec.ts
```diff
@@ -16,8 +16,8 @@ test('Open Trace from Explorer', async ({ page }) => {
 
 test('Open Trace from Trace Viewer', async ({ page }) => {
     await page.getByRole('tab', { name: 'Trace Viewer' }).locator('a').click();
-    await page.getByRole('button', { name: 'Opened Traces Section' }).hover();
-    await page.getByRole('button', { name: 'Open Trace' }).click();
+    await page.getByRole('button', { name: 'Welcome' }).hover();
+    await page.getByRole('button', { name: 'Open Folder' }).click();
```
bhufmann, line 20: I think `Open Trace` is till correct and should not be changed.

vscode-trace-extension/package.json
```diff
@@ -124,43 +124,56 @@
...
         }
       ]
     },
     "viewsWelcome": [
       {
-        "view": "traces",
-        "contents": "No traces found [learn more](https://www.eclipse.org/tracecompass/).\n[Add traces](command:nodeDependencies.addEntry)"
+        "view": "welcome",
+        "name": "Start Server",
+        "contents": "The server is offline. \n[Resume Trace Extension](command:vscode-trace-server.start-if-stopped)",
```
bhufmann, line 163: unfortunately, the `vscode-trace-extension` needs to be independent to the `vscode-trace-server` extension. There are users that don't use the `vscode-trace-server` extension and have their own way of starting the trace server. I suggest to look into the external API for starting a trace server. Right now a third-party extension (e.g. `vscode-trace-server`) can register their implementation to start the server. The current API requires a path to be passed. Maybe you can augment the API to add a default start api without passing the path. This should be optional.

vscode-trace-extension/package.json
```diff
@@ -124,43 +124,56 @@
...
+        "name": "Open Trace",
+        "contents": "There are currently no opened traces. \n[Open Folder](command:openedTraces.openTraceFolder)",
+        "when": "trace-explorer.noExperiments || !traceViewer.serverOn"
       },
       {
-        "view": "item properties",
-        "contents": "No properties to show"
+        "view": "welcome",
+        "name": "Documentation",
+        "contents": "To learn more about how to use the Trace Viewer Extension [read our documentation](https://github.com/eclipse-cdt-cloud/theia-trace-extension).",
```
bhufmann, line 175: This is great. Suggestion for a subsequent PR. I wonder if there is a need to make the link configurable so that adopters of the viewer can plug-in a link to their documentation.

vscode-trace-extension/package.json
bhufmann, line 163: `Trace Extension` is not the name that is used for the extension. The published name is `Trace Viewer` and it should be aligned with it: `Resume Trace Viewer`

vscode-trace-extension/package.json
```diff
@@ -124,43 +124,56 @@
...
+        "name": "Start Server",
+        "contents": "The server is offline. \n[Resume Trace Extension](command:vscode-trace-server.start-if-stopped)",
+        "when": "!traceViewer.serverOn"
       },
       {
-        "view": "views",
-        "contents": "No trace selected [learn more](https://www.eclipse.org/tracecompass/)"
+        "view": "welcome",
+        "name": "Open Trace",
+        "contents": "There are currently no opened traces. \n[Open Folder](command:openedTraces.openTraceFolder)",
```
bhufmann, line 169: `Open Trace`

vscode-trace-extension/package.json
bhufmann, line 175: Also remove `Extension` from the contents line: ... use the Trace Viewer [read documentation] ...

vscode-trace-extension/package.json
bhufmann, line 163: `The server is offline` should be `The trace server is offline`

vscode-trace-extension/package.json
bhufmann, line 175: The link should point to the `vscode-trace-extension` repository

bhufmann (changes requested):
Very nice improvement. Thank you very much. I have some comments though.

vscode-trace-extension/src/utils/backend-tsp-client-provider.ts
```diff
@@ -41,7 +41,27 @@ export const addTspClientChangeListener = (listenerFunction: (tspClient: TspClie
 export async function isUp(): Promise<boolean> {
     const health = await getTspClient().checkHealth();
     const status = health.getModel()?.status;
-    return health.isOk() && status === 'UP';
+    const serverIsUp = health.isOk() && status === 'UP';
+    vscode.commands.executeCommand('setContext', 'traceViewer.serverOn', serverIsUp);
+    if (serverIsUp) {
+        // If the server is up, check for opened experiments
+        areExperimentsOpened();
```
marcdumais-work, line 48: This, along with line 58 below in `areExperimentsOpened()`, which calls-back `isUp()`, seem to introduce a potential for an infinite loop, when the server is detected as being "up".

williamsyang-work: addressed

vscode-trace-extension/package.json
williamsyang-work: addressed

vscode-trace-extension/package.json
williamsyang-work: addressed

vscode-trace-extension/package.json
williamsyang-work: addressed

vscode-trace-extension/package.json
williamsyang-work: addressed

vscode-trace-extension/package.json
williamsyang-work: addressed

vscode-trace-extension/src/test/extension-test.spec.ts
williamsyang-work: addressed

vscode-trace-extension/package.json
williamsyang-work: Looking at the `trace-server-manager` code I see a way to de-couple the starting of a trace server with opening a trace.  However, I will implement this into a separate PR.

For now, I'm simply going to remove the Start Server welcome view from this PR.  The code for this `viewsWelcome` contribution will be simple enough to re-introduce in the future.

williamsyang-work:
There is currently a bug with the 'Open Trace' screen getting stuck after opening a trace in the local use-case.

image

vscode-trace-extension/package.json
williamsyang-work: After the `.vsix` file is built for an extension, it can be unzipped and the original package.json file can be modified in post.  This can be done using `jq` or running a node script from the shell.  I have done this to modify the default setting for server path with the vscode-trace-server extension.

vscode-trace-extension/src/utils/backend-tsp-client-provider.ts
```diff
@@ -41,7 +41,23 @@ export const addTspClientChangeListener = (listenerFunction: (tspClient: TspClie
 export async function isUp(): Promise<boolean> {
```
marcdumais-work, line 41: Since this is exported, and it's getting used more now, I think a more descriptive name would be good, like `isServerUp()` or `isTraceServerUp()`. Specially since the module name (`backend-tsp-client-provider`) does not clarify much, what exactly "is up".

vscode-trace-extension/src/utils/backend-tsp-client-provider.ts
```diff
@@ -41,7 +41,23 @@ export const addTspClientChangeListener = (listenerFunction: (tspClient: TspClie
...
     const health = await getTspClient().checkHealth();
     const status = health.getModel()?.status;
-    return health.isOk() && status === 'UP';
+    const serverIsUp = health.isOk() && status === 'UP';
+    vscode.commands.executeCommand('setContext', 'traceViewer.serverOn', serverIsUp);
```
marcdumais-work, line 45: Update the JSDoc to mention that the modified context?

marcdumais-work, line 45: For consistency with `TSP`, should we use `traceViewer.serverUp`?

vscode-trace-extension/src/utils/backend-tsp-client-provider.ts
```diff
@@ -41,7 +41,23 @@ export const addTspClientChangeListener = (listenerFunction: (tspClient: TspClie
...
+    return serverIsUp;
+}
+
+/**
+ * Checks for opened traces and updates the vscode context `trace-explorer.noExperiments`
+ * @returns noExperiments as boolean
+ */
+export async function areExperimentsOpened(): Promise<boolean> {
```
marcdumais-work, line 53: ATM, the returned boolean is not used. I am a bit disturbed by the reversal in meaning of the returned value vs function name. One potential way forward would be for this function to focus on updating the `noExperiment` context and rename it in consequence. e.g. `updateNoExperimentsContext()`

vscode-trace-extension/src/utils/backend-tsp-client-provider.ts
williamsyang-work: addressed

vscode-trace-extension/src/utils/backend-tsp-client-provider.ts
williamsyang-work: Yea this makes sense.  Addressed.

vscode-trace-extension/src/utils/backend-tsp-client-provider.ts
williamsyang-work: addressed

vscode-trace-extension/src/utils/backend-tsp-client-provider.ts
williamsyang-work: addressed

vscode-trace-extension/package.json
bhufmann: Thanks for the suggestion!

vscode-trace-extension/package.json
bhufmann: Thanks. Sounds good to split the PR.

bhufmann (changes requested):
Thanks for the updates. The code works well. One thing noticed though, when starting the trace server from command-line, then then there is now way to force refresh the trace explorer. Before it was possible to select the file explorer and then select the trace explorer to do a refresh. Now, this won't cause the trace explorer to  be populated with the opened traces. I think some user might start the server externally and for debugging of the server it would be good to support still that. Could you please look into that?

About the playwright tests, see [here](https://github.com/eclipse-cdt-cloud/vscode-trace-extension?tab=readme-ov-file#running-ui-tests) in the README on how to run it locally and debug.

williamsyang-work:
> Thanks for the updates. The code works well. One thing I noticed though, when starting the trace server from command-line, then then there is now way to force refresh the trace explorer. Before it was possible to select the file explorer and then select the trace explorer to do a refresh. Now, this won't cause the trace explorer to be populated with the opened traces. I think some user might start the server externally and for debugging of the server it would be good to support still that. Could you please look into that?
> 
> About the playwright tests, see [here](https://github.com/eclipse-cdt-cloud/vscode-trace-extension?tab=readme-ov-file#running-ui-tests) in the README on how to run it locally and debug.

I noticed this as well.  This is because with the new `viewsWelcome` there is no functionality to execute a function when the visibility changes.

I also noticed an edge-case where when you start a trace-server with no previously loaded traces, and simultaneously open a trace, the trace will load but the `viewsWelcome` will not go a way.  And, the only way to refresh the state is to open another trace.

I'm going to put this back into a draft for now.  The above case was avoidable with the additional welcome view of `Resume Trace Viewer` functionality.  I'll address your concerns with the ability to refresh the state, as well as this edge case, then I will re-open this PR.

vscode-trace-extension/src/test/extension-test.spec.ts
```diff
@@ -16,7 +16,7 @@ test('Open Trace from Explorer', async ({ page }) => {
 
 test('Open Trace from Trace Viewer', async ({ page }) => {
     await page.getByRole('tab', { name: 'Trace Viewer' }).locator('a').click();
-    await page.getByRole('button', { name: 'Opened Traces Section' }).hover();
+    await page.getByRole('heading', { name: 'TRACE VIEWER: WELCOME' }).hover();
```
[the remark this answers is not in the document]
bhufmann: > When I tested it locally I had to revert that change and it worked. What error did you notice that triggered this change?
> 
> Note, when running this test case, the first test case in this file has run before and there is already a trace on the server, and hence the trace explorer opens and not your welcome page.

I had a trace on the server before running the tests. Starting from an empty server, then I see that the welcome view is shown even if the first test case ran before which added a trace.

williamsyang-work:
I added a button that refreshes the state / context of the welcome view.  I could not find a way to refresh the state by navigating away and re-selecting. However, this seems like a common practice for extension development.

Two examples: GitLens and SourceControl:

gitlens-refresh

source-control-refresh

vscode-trace-extension/package.json
```diff
@@ -124,43 +129,50 @@
     },
     "views": {
       "trace-explorer": [
+        {
```
ngondalia, line 132: I have some general comments/concerns regarding this approach. I recently raised an issue (https://github.com/eclipse-cdt-cloud/theia-trace-extension/issues/1053) to handle single trace files along with trace folders. With that, i was hoping to achieve two things:

1) Have a separate button to handle trace files (this is simple to achieve with your approach)

2) Extensions have the ability to render either one or both of the buttons - this is useful for cases when users/extending extensions dont have a use case for one of the buttons (e.g. our users only ever need to inspect files and never folders) -- I am not sure if that is possible with this configuration as far as i know. Maybe it is possible to have different welcome views based on when clause context?

williamsyang-work: > I have some general comments/concerns regarding this approach. I recently raised an issue ([eclipse-cdt-cloud/theia-trace-extension#1053](https://github.com/eclipse-cdt-cloud/theia-trace-extension/issues/1053)) to handle single trace files along with trace folders. With that, i was hoping to achieve two things:
> 
> 1. Have a separate button to handle trace files (this is simple to achieve with your approach)
> 2. Extensions have the ability to render either one or both of the buttons - this is useful for cases when users/extending extensions dont have a use case for one of the buttons (e.g. our users only ever need to inspect files and never folders) -- I am not sure if that is possible with this configuration as far as i know. Maybe it is possible to have different welcome views based on when clause context?

1. I've added this functionality to this PR in a separate commit.  I think there is value in having this available by default, so just put it into the default view.
2. See this comment about modifying the contents of an extension's packate.json post-build: https://github.com/eclipse-cdt-cloud/vscode-trace-extension/pull/216#discussion_r1520809823.  I'm not sure if this helps you with your use-case, but its currently how I modify contents in order to suit custom needs.

williamsyang-work:
@bhufmann the problem with the UI test has been resolved.  Ready for re-review.

vscode-trace-extension/package.json
bhufmann: > 
>     2. Extensions have the ability to render either one or both of the buttons - this is useful for cases when users/extending extensions dont have a use case for one of the buttons (e.g. our users only ever need to inspect files and never folders) -- I am not sure if that is possible with this configuration as far as i know. Maybe it is possible to have different welcome views based on when clause context?

It's possible. In the packages.json you can have to entries for open directory and open files. Then there can be context flag added in the when-clause to show it or not. 
```json
"viewsWelcome": [
      {
        "view": "welcome",
        "name": "Open Trace Directory",
        "contents": "There are currently no opened traces.\n[Open Trace Folder](command:openedTraces.openTraceFolder)",
        "when": "trace-explorer.noExperiments || !traceViewer.serverUp"
      },
      {
        "view": "welcome",
        "name": "Open Trace File",
        "contents": "There are currently no opened traces.\n[Open Trace File](command:openedTraces.openTraceFile)",
        "when": "(trace-explorer.noExperiments || !traceViewer.serverUp) && trace-explorer.hasFileOpen"
      },      
      {
        "view": "welcome",
        "name": "Documentation",
        "contents": "To learn more about how to use the Trace Viewer [read our documentation](https://github.com/eclipse-cdt-cloud/vscode-trace-extension).",
        "when": "trace-explorer.noExperiments || !traceViewer.serverUp"
      }
    ],
```

This flag can be set programmatically from any extension.
```javascript
vscode.commands.executeCommand('setContext', 'trace-explorer.hasFileOpen', true);
```

@ngondalia  is that what you were looking for?

vscode-trace-extension/package.json
```diff
@@ -169,6 +181,11 @@
           "command": "openedTraces.openTraceFolder",
           "when": "view == traceExplorer.openedTracesView && !trace-explorer.noExperiments",
           "group": "navigation"
+        },
+        {
+          "command": "trace-explorer.refreshContext",
```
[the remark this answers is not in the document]
williamsyang-work: If you look at the last commit on line 211 I added that functionality in.  So, on application startup, there is an initial execution of `trace-explorer.refreshContext`

vscode-trace-extension/package.json
```diff
@@ -124,43 +129,50 @@
...
         }
       ]
     },
     "viewsWelcome": [
       {
-        "view": "traces",
-        "contents": "No traces found [learn more](https://www.eclipse.org/tracecompass/).\n[Add traces](command:nodeDependencies.addEntry)"
+        "view": "welcome",
+        "name": "Open Trace",
+        "contents": "There are currently no opened traces.\n[Open Trace Folder](command:openedTraces.openTraceFolder)\n[Open Trace File](command:openedTraces.openTraceFile)",
```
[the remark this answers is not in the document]
williamsyang-work: Sounds good.  It has been removed.

vscode-trace-extension/package.json
ngondalia: > > 1. Have a separate button to handle trace files (this is simple to achieve with your approach)
> 
> I've added this functionality to this PR in a separate commit. I think there is value in having this available by default, so just put it into the default view.
> 
> > 2. Extensions have the ability to render either one or both of the buttons - this is useful for cases when users/extending extensions dont have a use case for one of the buttons (e.g. our users only ever need to inspect files and never folders) -- I am not sure if that is possible with this configuration as far as i know. Maybe it is possible to have different welcome views based on when clause context?
> 
> See this comment about modifying the contents of an extension's packate.json post-build: [#216 (comment)](https://github.com/eclipse-cdt-cloud/vscode-trace-extension/pull/216#discussion_r1520809823). I'm not sure if this helps you with your use-case, but its currently how I modify contents in order to suit custom needs.

- That wouldn't work for our case becasue we are not publishing our own version of the extension - we will be relying on the marketplace version, so its more preferrable to do it through API/contexts

ngondalia: > > ```
> > 2. Extensions have the ability to render either one or both of the buttons - this is useful for cases when users/extending extensions dont have a use case for one of the buttons (e.g. our users only ever need to inspect files and never folders) -- I am not sure if that is possible with this configuration as far as i know. Maybe it is possible to have different welcome views based on when clause context?
> > ```
> 
> It's possible. In the packages.json you can have to entries for open directory and open files. Then there can be context flag (`hasFileOpen`) added that is checked in the when-clause. The default value would be `true`.
> 
> ```json
> "viewsWelcome": [
>       {
>         "view": "welcome",
>         "name": "Open Trace Directory",
>         "contents": "There are currently no opened traces.\n[Open Trace Folder](command:openedTraces.openTraceFolder)",
>         "when": "trace-explorer.noExperiments || !traceViewer.serverUp"
>       },
>       {
>         "view": "welcome",
>         "name": "Open Trace File",
>         "contents": "There are currently no opened traces.\n[Open Trace File](command:openedTraces.openTraceFile)",
>         "when": "(trace-explorer.noExperiments || !traceViewer.serverUp) && trace-explorer.hasFileOpen"
>       },      
>       {
>         "view": "welcome",
>         "name": "Documentation",
>         "contents": "To learn more about how to use the Trace Viewer [read our documentation](https://github.com/eclipse-cdt-cloud/vscode-trace-extension).",
>         "when": "trace-explorer.noExperiments || !traceViewer.serverUp"
>       }
>     ],
> ```
> 
> This flag can be set programmatically from any extension.
> 
> ```js
> vscode.commands.executeCommand('setContext', 'trace-explorer.hasFileOpen', false);
> ```
> 
> @ngondalia is that what you were looking for?

- Yeah, that is exactly what i was thinking! The only thing i would suggest is to break it down further to split the two buttons and that can have its own when context clauses:

```
"viewsWelcome": [
      {
        "view": "welcome",
        "name": "Open Trace Directory",
        "contents": "There are currently no opened traces.\n[Open Trace Folder](command:openedTraces.openTraceFolder)",
        "when": "trace-explorer.noExperiments || !traceViewer.serverUp"
      },
      {
        "view": "welcome",
        "name": "Open Trace File",
        "contents": "There are currently no opened traces.\n[Open Trace File](command:openedTraces.openTraceFile)",
        "when": "(trace-explorer.noExperiments || !traceViewer.serverUp) && trace-explorer.hasFileOpen"
      },      
      {
        "view": "welcome",
        "name": "Documentation",
        "contents": "To learn more about how to use the Trace Viewer [read our documentation](https://github.com/eclipse-cdt-cloud/vscode-trace-extension).",
        "when": "trace-explorer.noExperiments || !traceViewer.serverUp"
      }
    ]
```

bhufmann (dismissed):
It looks good to me. Thanks for the UX improvement. 

@ngondalia We'd like to merge it. The `Open Trace File` would have to be added to the welcome page since it's not re-using the PlaceHolderWidget of the traceviewer-react-components. Thanks, Bernd

ngondalia:
> It looks good to me. Thanks for the UX improvement.
> 
> @ngondalia We'd like to merge it. The `Open Trace File` would have to be added to the welcome page since it's not re-using the PlaceHolderWidget of the traceviewer-react-components. Thanks, Bernd

No issues from my end! The "Open Trace File" can be addressed in a separate PR.

vscode-trace-extension/package.json
```diff
@@ -124,43 +129,50 @@
...
       {
-        "view": "views",
-        "contents": "No trace selected [learn more](https://www.eclipse.org/tracecompass/)"
-      },
-      {
-        "view": "item properties",
-        "contents": "No properties to show"
+        "view": "welcome",
+        "name": "Documentation",
+        "contents": "To learn more about how to use the Trace Viewer [read our documentation](https://github.com/eclipse-cdt-cloud/vscode-trace-extension).",
```
marcdumais-work, line 174: I think it would be more appropriate to point to the [extension's README](https://github.com/eclipse-cdt-cloud/vscode-trace-extension/blob/master/vscode-trace-extension/README.md) vs the repo root

vscode-trace-extension/package.json
```diff
@@ -89,6 +89,11 @@
         "title": "Open Trace",
         "icon": "$(new-folder)"
       },
+      {
+        "command": "trace-explorer.refreshContext",
+        "title": "Refresh",
```
marcdumais-work, line 94: image

The name should be more descriptive. Something like "Refresh Trace Server"?

bhufmann: I think it should be `Refresh Trace Viewer` or `Refresh Trace Explorer` because it's the front-end view that is being refreshed using the back-end sever.

vscode-trace-extension/src/utils/backend-tsp-client-provider.ts
```diff
@@ -36,12 +36,29 @@ export const addTspClientChangeListener = (listenerFunction: (tspClient: TspClie
...
+ * @returns noExperiments as boolean
+ */
+export async function updateNoExperimentsContext(): Promise<boolean> {
+    const response = await getTspClient().fetchExperiments();
+    if (!response.isOk()) {
+        return false;
+    }
+    const noExperiments = !response.getModel()?.length;
+    vscode.commands.executeCommand('setContext', 'trace-explorer.noExperiments', noExperiments);
+    return noExperiments;
```
marcdumais-work, line 61: Return value not used anywhere? Remove if not needed.

vscode-trace-extension/package.json
williamsyang-work: addressed

vscode-trace-extension/package.json
williamsyang-work: addressed

vscode-trace-extension/src/utils/backend-tsp-client-provider.ts
williamsyang-work: addressed

vscode-trace-extension/package.json
marcdumais-work: I still see the original URL, pointing to the repo root README (which is more geared towards developers)

vscode-trace-extension/src/utils/backend-tsp-client-provider.ts
```diff
@@ -36,12 +36,29 @@ export const addTspClientChangeListener = (listenerFunction: (tspClient: TspClie
...
     const status = health.getModel()?.status;
-    return health.isOk() && status === 'UP';
+    const serverIsUp = health.isOk() && status === 'UP';
+    vscode.commands.executeCommand('setContext', 'traceViewer.serverUp', serverIsUp);
+    return serverIsUp;
+}
+
+/**
+ * Checks for opened traces and updates the vscode context `trace-explorer.noExperiments`
+ * @returns noExperiments as boolean
```
marcdumais-work, line 52: no longer in-sync

marcdumais-work: looks good now

marcdumais-work (approved):
LGTM, thanks for this nice contribution!

bhufmann (approved):
Looks good to me. Thanks for the UX improvement.

PR #221, opened 2024-03-21
Handling of Files with Open Trace Button

ngondalia:
This commit adds a handler that manages and detects resource type dialog that should be opened when opening a trace. It provides the ability to handle one or both file and folder trace types. If both is handled, a quick pick is shown to select which type of resource dialog should be opened. The functionality to control which trace type should be handle is provided through
the external API.

### DIFF
--- a/vscode-trace-extension/src/trace-explorer/trace-utils.ts
+++ b/vscode-trace-extension/src/trace-explorer/trace-utils.ts
@@ -35,11 +35,11 @@ export const zoomHandler = (hasZoomedIn: boolean): void => {
     TraceViewerPanel.zoomOnCurrent(hasZoomedIn);
 };
 
-export const openDialog = async (): Promise<vscode.Uri | undefined> => {
+export const openDialog = async (selectFiles = false): Promise<vscode.Uri | undefined> => {
     const props: vscode.OpenDialogOptions = {
-        title: 'Open Trace',
-        canSelectFolders: true,
-        canSelectFiles: false,
+        title: selectFiles ? 'Open Trace File' : 'Open Trace Folder',
+        canSelectFolders: !selectFiles,
+        canSelectFiles: selectFiles,
         canSelectMany: false
     };
     let traceURI = undefined;
--- a/vscode-trace-extension/src/utils/trace-explorer-resource-type-handler.ts
+++ b/vscode-trace-extension/src/utils/trace-explorer-resource-type-handler.ts
@@ -1,0 +1,84 @@
+/***************************************************************************************
+ * Copyright (c) 2024 BlackBerry Limited and others.
+ *
+ * Licensed under the MIT license. See LICENSE file in the project root for details.
+ ***************************************************************************************/
+import * as vscode from 'vscode';
+
+export type ResourceType = 'File' | 'Folder';
+export type ResourceTypeQuickPickItem = vscode.QuickPickItem & { type: ResourceType };
+
+export class TraceExplorerResourceTypeHandler {
+    private static instance: TraceExplorerResourceTypeHandler;
+    private doHandleFiles = true;
+    private doHandleFolders = true;
+    private quickpickItems: ResourceTypeQuickPickItem[] = [
+        { label: 'File', type: 'File' },
+        { label: 'Folder', type: 'Folder' }
+    ];
+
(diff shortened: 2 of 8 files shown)


ngondalia:
@bhufmann This is the follow up commit for handling of trace files with "Open Trace" button. 

In this commit, i decided to use a quick pick to select if it should be a file or folder dialog when both trace types are supported.
When only one is set, the TraceExplorerResourceTypeHandler detects which one is set and shows the appropriate dialog.

The other option is to have two separate buttons hooked to two separate commands that can be conditionally rendered (which we discussed in #216 ). 

I don't have a preference to which option, but let me know if you do and i can change whatever is needed.

Thanks! :)

bhufmann:
@ngondalia  I'm going to review this PR. Just a heads-up, I just merged a code clean-up patch #223 that causes merge conflicts (e.g. I renamed trace-tree.ts) with PR. You'll have to rebase and resolve the merge conflicts.

ngondalia:
> @ngondalia I'm going to review this PR. Just a heads-up, I just merged a code clean-up patch #223 that causes merge conflicts (e.g. I renamed trace-tree.ts) with PR. You'll have to rebase and resolve the merge conflicts.

Rebased!

bhufmann (changes requested):
It looks good to me. I ran it by some of my comments and we ok with this approach. I'd like to @hriday-panchasara to comment as well, since he was looking into that feature, too.

I have on comment though to fix.

vscode-trace-extension/src/utils/trace-explorer-resource-type-handler.ts
```diff
@@ -0,0 +1,82 @@
...
+export class TraceExplorerResourceTypeHandler {
+    private static instance: TraceExplorerResourceTypeHandler;
+    private doHandleFiles = true;
+    private doHandleFolders = true;
+    private quickpickItems: ResourceTypeQuickPickItem[] = [
+        { label: 'File', type: 'File' },
+        { label: 'Folder', type: 'Folder' }
+    ];
+
+    private constructor() {}
```
hriday-panchasara, line 20: Getting ESLint "no-empty-function" error. You can resolve this by adding an empty comment on purpose

` private constructor() { /** Empty constructor */ }`

vscode-trace-extension/src/trace-explorer/trace-utils.ts
```diff
@@ -35,11 +35,11 @@ export const zoomHandler = (hasZoomedIn: boolean): void => {
     TraceViewerPanel.zoomOnCurrent(hasZoomedIn);
 };
 
-export const openDialog = async (): Promise<vscode.Uri | undefined> => {
+export const openDialog = async (selectFiles: boolean = false): Promise<vscode.Uri | undefined> => {
```
hriday-panchasara, line 38: Type boolean inferred from boolean literal. Can change to:
`selectFiles = false`

hriday-panchasara (approved):
This approach works for me as well. But we should adjust the Theia implementation to use a dropdown pick or quick pick in order to have some consistency.

vscode-trace-extension/src/trace-explorer/trace-utils.ts
ngondalia: Fixed!

vscode-trace-extension/src/utils/trace-explorer-resource-type-handler.ts
ngondalia: Ok, fixed!

ngondalia:
> This approach works for me as well. But we should adjust the Theia implementation to use a dropdown pick or quick pick in order to have some consistency.

Yeah - I agree!

bhufmann (approved):
Looks good to me. Thanks for the contribution!

bhufmann:
> > This approach works for me as well. But we should adjust the Theia implementation to use a dropdown pick or quick pick in order to have some consistency.
> 
> Yeah - I agree!

@ngondalia please let me know if you going to work on this update. Theia has the `QuickPickService` that can be used for that (`@theia/core/lib/common/quick-pick-service`).

hriday-panchasara (approved):
Thanks for the contribution!

PR #222, opened 2024-03-22
AbstractTraceExplorerProvider - Code Cleanup

williamsyang-work:
Factors out common logic from all WebviewViewProviders into a new
AbstractTraceExplorerProvider abstract class.  This allows for shared
functionality between WebviewViewProviders to be modified in a single
place instead of each individual file.  DRYs out the code.


MatthewKhouzam (approved):
LGTM

bhufmann (approved):
Looks good to me. Thanks for the clean-up!

PR #223, opened 2024-03-25
Remove obsolete vscode.TreeDataProvider implementation

bhufmann:
This code was a left-over when the proof-of-concept for the `vscode-trace-extension` was started and both the `open traces view` and `analyses view` were implemented as tree views. After the initial implementation, the code was changed to use the reusable widgets from the `traceviewer-react-components` of the `theia-trace-extension` repository.


marcdumais-work (approved):
LGTM - nice bit of Spring cleaning!

PR #226, opened 2024-03-26
Add Semi-Opaque 'Server Offline' UI Blocker in TraceViewerPanels

williamsyang-work:
Add UI blocking overlay when server is offline.  This is to prevent the pending-requests issue in the remote use case and also prevents breaking ui states in the local use case.

Modifies the TraceServerConnectionStatusService to send a vscode message informing all active webviews (views and panels) of the change in server status. Adds a public method in TraceViewerPanel class that sends a vscode message to all active panels.

### DIFF
--- a/vscode-trace-extension/src/utils/trace-server-status.ts
+++ b/vscode-trace-extension/src/utils/trace-server-status.ts
@@ -1,13 +1,15 @@
 import { ThemeColor, StatusBarItem } from 'vscode';
 import { isTraceServerUp } from './backend-tsp-client-provider';
 import * as vscode from 'vscode';
+import { VSCODE_MESSAGES } from 'vscode-trace-common/lib/messages/vscode-message-manager';
+import { TraceViewerPanel } from '../trace-viewer-panel/trace-viewer-webview-panel';
 
 export class TraceServerConnectionStatusService {
-    private statusBarItem: StatusBarItem;
+    private _status = false;
 
-    public constructor(statusBarItem: StatusBarItem) {
-        this.statusBarItem = statusBarItem;
-        this.statusBarItem.hide();
+    public constructor(private _statusBarItem: StatusBarItem) {
+        _statusBarItem.hide();
+        this.checkAndUpdateServerStatus();
     }
 
     public checkAndUpdateServerStatus = async (): Promise<void> => {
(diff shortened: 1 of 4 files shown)


bhufmann (commented):
@williamsyang-work #230 fixes some issues with the refresh of the views after server start/stop/config-change. It impacts this PR. One thing I changed is that when a component knows that the server is down, there is no need to check if server is down. There will be merge conflicts. Could you please have look if it still works after rebase of my updates? Thanks!

williamsyang-work:
> @williamsyang-work #230 fixes some issues with the refresh of the views after server start/stop/config-change. It impacts this PR. One thing I changed is that when a component knows that the server is down, there is no need to check if server is down. There will be merge conflicts. Could you please have look if it still works after rebase of my updates? Thanks!

10-4

bhufmann:
@williamsyang-work PR #230 has been merged. Please rebase this PR to that version. I'll review it once it's available. Thanks!

williamsyang-work:
> @williamsyang-work PR #230 has been merged. Please rebase this PR to that version. I'll review it once it's available. Thanks!

PR has been rebased!

vscode-trace-webviews/src/trace-viewer/index.css
```diff
@@ -10,3 +10,26 @@ body {
     width: 100%;
     position: absolute;
 }
+
+.overlay {
```
bhufmann, line 14: When using the dark theme, then the background is barely noticeable. Could you please make it more visible in dark theme?

vscode-opaque-overlay

bhufmann (changes requested):
It works well. One comment about dark theme below.

Also, do you plan to disable the toolbar actions (for example zoom, home buttons) as well when server is down? They also interact with the server. Please let me know.

vscode-trace-webviews/src/trace-viewer/index.css
williamsyang-work: The opacity has been increased 0.25 => 0.5

bhufmann: Looks better now. Thanks!

bhufmann:
> Also, do you plan to disable the toolbar actions (for example zoom, home buttons) as well when server is down? They also interact with the server. Please let me know.

Thinking more about it's out of scope of this PR and can be addressed in a later PR.

bhufmann (approved):
LGTM. Thanks!

williamsyang-work:
> > Also, do you plan to disable the toolbar actions (for example zoom, home buttons) as well when server is down? They also interact with the server. Please let me know.
> 
> Thinking more about it's out of scope of this PR and can be addressed in a later PR.

This is a great idea that I didn't think of.

vscode-trace-extension/src/utils/trace-server-status.ts
```diff
@@ -16,20 +19,34 @@ export class TraceServerConnectionStatusService {
...
+        }
+        this._status = status;
+
         await vscode.commands.executeCommand('setContext', 'traceViewer.serverUp', status);
+
+        const command = VSCODE_MESSAGES.CONNECTION_STATUS;
+        const data = this._status.toString();
+
+        traceExtensionWebviewManager.getAllActiveWebviews().forEach(_view => {
+            _view.webview.postMessage({ command, data });
```
marcdumais-work, line 33: I'm wondering if this is still required, after the recent commit where webviews are disposed automatically (https://github.com/eclipse-cdt-cloud/vscode-trace-extension/commit/77bf4f03c44677685bcb4ae92003d7f43f3a0be8)?

Debugging through the code, it seems that on trace server shutdown, the webviews are already disposed by the time this is executed, and on trace server startup, the webviews are freshly created and may not need this update?

williamsyang-work: I agree.  I added this to be thorough and send to all webviews, but
a) the explorer webviews should be disposed when the welcome screen is rendered (what you said)
b) the webviews don't have any listeners for this event.

I went ahead and removed this for now.  It can be re-added in the future if necessary.

bhufmann (approved):
Still looks good to me after latest update.

marcdumais-work (approved):
LGTM - another really nice contribution, thanks @williamsyang-work!

Issue #228, opened 2024-03-27
Welcome Page not rendered after trace server stopped indication (intermittent)


bhufmann:
When stopping the trace server using the [vscode-trace-server](https://github.com/eclipse-cdt-cloud/vscode-trace-server) stop command then the Trace Explorer stays and the welcome page is not rendered, nor the status bar is not updated. One has to press the refresh button to refresh the view. Note that this is intermittent and doesn't happen all the time. 

![vscode-no-refresh-at-stop](https://github.com/eclipse-cdt-cloud/vscode-trace-extension/assets/413424/5f0719bf-0971-44db-9765-8564cee239af)

PR #230, opened 2024-03-27
Fix trace explorer / welcome page refresh corner cases

### LINKED ISSUE #227: Trace Explorer not refreshed after first trace server started indication
When starting the trace server using the [vscode-trace-server](https://github.com/eclipse-cdt-cloud/vscode-trace-server) start command then the Trace Explorer is not refreshed and the welcome page stays. One has to press the refresh button to refresh the view. Note that this only happens the first time after opening the trace viewer.

vscode-no-refresh-at-start

### LINKED ISSUE #229: Trace Explorer/Welcome page not refreshed when updating the server URL (corner case)
When updating the server URL preference to connect to a different trace server, the Trace Explorer is not rendered if the origin trace server doesn't have any traces or server is not started. Hence after the update, the welcome page stays. The refresh works fine when the original server has traces and the trace explorer is rendered. One has to press the refresh button to refresh the view. 

vscode-no-refresh-at-change-url

bhufmann:
This PR consists of multiple commits that fixes refresh issues #227, #228 and #229. 

1. Add default dispose() method to AbstractTraceExplorerProvider
Call it from the webview dispose handler. Sub-class can add their own implementation and call super implementation. This avoids that the class is using the webview instance after it is disposed.

2. Ignore connection status listener notification in constructor
The TSP client implementation notifies clients when the method `addConnectionStatusListener()` is called. This can lead to incorrect state in on client side.

3. Add method to set server status in TraceServerStatusService 
This method set the status bar and the `traceViewer.serverUp` context that is used to decide whether the trace explorer or the welcome view should be rendered. This method doesn't check the backend to get status in comparison to `checkAndUpdateServerStatus`. It's useful when it's known that the server is down or up (e.g. after start notification). In this case, it is not necessary to check the backend again. Don't set the status in the method `isTraceServerUp()` because it should not have side-effects that are not expected.

Initialize and refresh flags `traceViewer.serverUp` and `trace-explorer.noExperiments` to ensure the correct view is rendered,
i.e. Trace Explorer or Welcome view, when certain events happen (e.g. server started or stopped). Fixes #227 and #228

4. Refresh trace viewer after server URL preference changes
For the case that with current URL the welcome view is rendered due to no experiments on server, or the server is down, this will render the trace explorer (or welcome view) after a server URL change. Without calling refresh, the welcome view stays, even if the other server has experiments. Fixes #229


bhufmann:
Video showing fix for #227:

vscode-refresh-at-start

bhufmann:
Video showing fix for #228 

vscode-refresh-at-stop

bhufmann:
Video showing fix for #229 

vscode-refresh-at-change-url

bhufmann:
@williamsyang-work I found some corner cases for the refresh of the trace-explorer/welcome-view that this PR addresses. Please let me know if have any comments for this PR.

vscode-trace-common/src/client/tsp-client-provider-impl.ts
```diff
@@ -10,13 +10,21 @@ export class TspClientProvider implements ITspClientProvider {
...
         this.updateClients();
-        RestClient.addConnectionStatusListener(status => this._signalHandler?.notifyConnection(status));
+
+        RestClient.addConnectionStatusListener(status => {
+            // Ignore the first update that is sent when calling addConnectionStatusListener
+            if (this._initialized) {
+                this._signalHandler?.notifyConnection(status);
+            }
+        });
+        this._initialized = true;
```
williamsyang-work, line 27: Similar pattern to this: https://github.com/eclipse-cdt-cloud/vscode-trace-extension/pull/216/commits/f565c231cd9fd15aa0ab6932ec83a4a2dc96d651.

I wonder if there's a cleaner way to do this.

williamsyang-work:
Thanks for the assist on this @bhufmann :)

vscode-trace-common/src/client/tsp-client-provider-impl.ts
bhufmann: I'm not sure what you mean. What do you suggest to make it cleaner? Please let me know

marcdumais-work (approved):
LGTM - Very nice improvements for several corner-cases! Not only the views refresh correctly but it happens quicker now, without that little delay that was there in some cases previously. I confirmed that that 3 linked issues now seem resolved.

marcdumais-work:
@bhufmann FYI, I noticed another corner case, that was already present before this PR while testing: https://github.com/eclipse-cdt-cloud/vscode-trace-extension/issues/236

vscode-trace-common/src/client/tsp-client-provider-impl.ts
bhufmann: Ok, I think I understand how to change it. I'll update the PR (including a rebase).

bhufmann:
> @bhufmann FYI, I noticed another corner case, that was already present before this PR, while testing: #236

Thanks for providing this. I think that should be handled outside this PR in dedicated PR to fix #236.

PR #231, opened 2024-03-27
Make playwright tests more stable

bhufmann:
- Run tests serial: There is only one server and test can interfere each other
- Wait for Open Traces section to be visible
- Increase timeouts due slow opening of traces and switching views


bhufmann:
#221 fixes seems to fix the intermittent failure. Closing this one since it's not needed.

bhufmann:
Not need at the moment

bhufmann:
Not need at the moment

PR #234, opened 2024-04-04
Fixed message passing to webviews via AbstractTraceExplorerProvider

ngondalia:
Fixed bug in postMessagetoWebview method of AbstractTraceExplorerProvider. The method returned on a valid command causing some views to break (Item Properties view)


hriday-panchasara (approved):
LGTM, thanks!

PR #239, opened 2024-04-12
Fix handling of state in doHandleOutputAddedMessage

ngondalia:
This method is doing a check on the state to see if output descriptor already exists. In the case it is not found, it had a blocking call before updating the state. This resulted in subsequent signals received directly after the first one, to handle the same output descriptor again.

The outcome of this was that there would be descriptors with the same key being rendered causing errors.


bhufmann (approved):
Looks good to me. Thanks!

Issue #238, opened 2024-04-11
Endless switching of trace viewer tabs and selected traces after trace server start


bhufmann:
Having multiple traces open, then stop and start trace server, plus selecting tabs and traces after the trace server is finished starting, the view gets into the state that the tabs seems to be selected continuously. This can only be stopped by closing all traces but one. After that the available views are still updated for a while. 

It's looks like that there are signals sent internally in the trace viewer and they never stop being sent.

![vscode-samba-dancing](https://github.com/eclipse-cdt-cloud/vscode-trace-extension/assets/413424/e7d721f5-5d8e-41f6-814c-d56d05325142)

marcdumais-work:
I think maybe I have stumbled on an even simpler way to reproduce this issue. I will confirm and if so, capture a video.

PR #240, opened 2024-04-15
Handle multiple experiments with the same name

ngondalia:
Create panel after server instantiates the experiment and tie the panel name to the experiment name

Fixes the constant switching of tabs when multiple experiments opened with the same name

### DIFF
--- a/vscode-trace-extension/src/trace-explorer/trace-utils.ts
+++ b/vscode-trace-extension/src/trace-explorer/trace-utils.ts
@@ -69,16 +69,13 @@ export const fileHandler =
 
                 const filePath: string = resolvedTraceURI.fsPath;
                 if (!filePath) {
-                    traceLogger.addLogMessage(
-                        'Cannot open trace: could not retrieve path from URI for trace ' + resolvedTraceURI,
-                        fileHandler.name
+                    traceLogger.showError(
+                        'Cannot open trace: could not retrieve path from URI for trace ' + resolvedTraceURI
                     );
                     return;
                 }
 
                 const name = path.basename(filePath);
-                const panel = TraceViewerPanel.createOrShow(context.extensionUri, name, undefined);
-
                 progress.report({ message: ProgressMessages.FINDING_TRACES, increment: 10 });
                 /*
                  * TODO: use backend service to find traces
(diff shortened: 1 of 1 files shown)


ngondalia:
> light2.mp4

This is the issue that this commit addresses

vscode-trace-extension/src/trace-explorer/trace-utils.ts
```diff
@@ -114,29 +107,29 @@ export const fileHandler =
...
-                        traceLogger.addLogMessage('Failed to open trace: ' + traceName, fileHandler.name);
-                        traceLogger.addLogMessage(
-                            'There may be an issue with the server or the trace is invalid.',
-                            fileHandler.name
+                        traceLogger.showError(
+                            'Failed to open trace: ' +
+                                traceName +
+                                '. There may be an issue with the server or the trace is invalid.'
                         );
+                        return;
```
bhufmann, line 115: If returning here no experiment is created with a subset of traces. The idea was that for a list of traces pathes in tracesArray, each trace is posted to the server. If not successful, it continues with the next trace in the array. If one or more are successfully posted, the experiment is created with these traces. Below there is a warning message created to the user of failed traces. If no trace was successfully posted the no experiment is created. 

I think we should remove the return here

bhufmann (changes requested):
The change makes sense to open the panel with the correct name. 

I was not able to reproduce the flickering. I guess it's timing related. I got hopeful that it would fix the flickering reported in #238, but this update doesn't address it. The Trace Compass team will investigate that issue in a separate effort.

vscode-trace-extension/src/trace-explorer/trace-utils.ts
ngondalia: My mistake - fixed!

ngondalia:
> I was not able to reproduce the flickering. I guess it's timing related. I got hopeful that it would fix the flickering reported in #238, but this update doesn't address it. The Trace Compass team will investigate that issue in a separate effort.

It doesnt always happen - it depends on what you click after opening the second trace. For me, its always reproducible if i click the newly opened trace (2nd trace with same name) in Opened Traces view.

bhufmann:
> > I was not able to reproduce the flickering. I guess it's timing related. I got hopeful that it would fix the flickering reported in #238, but this update doesn't address it. The Trace Compass team will investigate that issue in a separate effort.
> 
> It doesnt always happen - it depends on what you click after opening the second trace. For me, its always reproducible if i click the newly opened trace (2nd trace with same name) in Opened Traces view.

I was able to reproduce it now. Thanks for the steps.

bhufmann (approved):
Looks good to me. Thanks!

PR #241, opened 2024-04-15
[ci] harmonize release doc, add dhelper scripts

marcdumais-work:
Make this repo more similar to our other cdt-cloud repos:
- prevent triggering release-related jobs on forks
- document release process in root README
- add package.json scripts to help create the release commit and step version

Also, pre-format the various `package.json` and `lerna.json` files, as per lerna expectations. 
- indent using 4 spaces
- ordered dependencies

This is the format lerna enforces. Doing it now means that when we eventually use lerna to step the package.json versions, only the version change will be reflected in the result.

_Note: because of the many indentation-related changes, it might be easier to review using the "hide whitespace" option, in the "files changes" panel._

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -95,7 +95,7 @@ jobs:
         os: [ubuntu-latest]
         node-version: [18]
     # Only execute when the trigger was a tag (new release)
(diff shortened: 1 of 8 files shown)


bhufmann (approved):
Looks good to me. Thanks!

marcdumais-work:
Thanks for the review!

PR #248, opened 2024-05-06
Handle new Item Properties Signal Payload

### LINKED ISSUE #202: Item properties do not persist when hidden and re-opened.
Need a way to keep track of the webview's state on the provider or through VSCode's API.

ngondalia:
This commit addresses the persistence of item properties and handling of the new ItemPropertiesSignalPayload.

Fixes #202


ngondalia:
Depends on https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/1075

PR #249, opened 2024-05-30
Move 'Open with Trace Viewer' option lower

williamsyang-work:
Moves it to a lower / more appropriate part of the context menu.

Before:
image

After:
image


marcdumais-work (approved):
LGTM, thanks @williamsyang-work !

PR #250, opened 2024-06-10
Fix endless switching between experiment panels

### LINKED ISSUE #238: Endless switching of trace viewer tabs and selected traces after trace server start

bhufmann:
This happens when opening multiple traces and manually switching newly opened panels.

- The main fix is to avoid calling panel.reveal() when the trace-viewer-webview-panel is processing the EXPERIMENT_SELECTED signal. This call is not only unecessary but will also interfere with ongoing propagation of TRACEVIEWERTAB_ACTIVATED and EXPERIMENT_SELECTED signals.

- Only call setExepriment() on the panel instance if it was created new, hence it was reopened, and don't call on the panel if the panel is already open. For that a new public static method getExistingPanel was added to the TraceViewerPanel class.

- Only call fireExperimentSelectedSignal() right after calling fireTraceViewerTabActivatedSignal() if the open-traces webview is not visible. If this webview is visible, then the webview component ReactOpenTracesWidget will send it to update the available-views view. This is now done in a central place where the
TraceExplorerOpenedTracesViewProvider handles the
TRACEVIEWERTAB_ACTIVATED signal. This avoids unnecessary sending and processing of the EXPERIMENT_SELECTED signal.

Fixes #238

### DIFF
--- a/vscode-trace-extension/src/trace-explorer/opened-traces/trace-explorer-opened-traces-webview-provider.ts
+++ b/vscode-trace-extension/src/trace-explorer/opened-traces/trace-explorer-opened-traces-webview-provider.ts
@@ -40,11 +40,19 @@ export class TraceExplorerOpenedTracesViewProvider extends AbstractTraceExplorer
     }
 
     protected doHandleTracesWidgetActivatedSignal(experiment: Experiment): void {
-        if (this._view && experiment) {
+        if (!experiment) {
+            return;
+        }
+        if (this._view) {
             this._selectedExperiment = experiment;
             const wrapper: string = JSONBig.stringify(experiment);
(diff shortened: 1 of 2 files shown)


marcdumais-work (approved):
LGTM, thanks Bernd! 

I've tested a bit around the changes and everything still seems to work well.

PR #252, opened 2024-06-10
Fix exception in TraceExplorerItemPropertiesProvider

bhufmann:
The PR fixes the following exception:
TypeError: Cannot read properties of undefined (reading 'UUID') at TraceExplorerItemPropertiesProvider.handleExperimentChanged


marcdumais-work (approved):
LGTM - thanks Bernd! 

I confirmed the exception is gone, when closing a trace using "remove trace from trace viewer" "x" button.

PR #262, opened 2024-08-19
Upgrade to traceviewer* packages to v0.2.6, that include upgraded ag-grid

marcdumais-work:
The new traceviewer packages contain the upgraded ag-grid packages, that removes several high-level vulnerabilities.
    
Also, in this repo here, we now only need the ag-grid styles package, "@ag-grid-community/styles" - the rest of ag-grid being obtained from `traceviewer-react-components`. Note that `ag-grid` now distributes themes containing both dark and light in the same .css file.

---

Performed a `yarn upgrade`: we went from 5 to 2 known vulnerabilities, as per "yarn audit":

Before:
5 vulnerabilities found - Packages audited: 926
Severity: 5 Moderate

After:
2 vulnerabilities found - Packages audited: 981
Severity: 2 Moderate

### DIFF
--- a/configs/license-check-config.json
+++ b/configs/license-check-config.json
@@ -4,5 +4,6 @@
     "inputFile": "yarn.lock",
     "batch": 50,
     "timeout": 200,
-    "summary": "license-check-summary.txt"
+    "summary": "license-check-summary.txt",
+    "exclusions": "configs/license-check-exclusions.json"
 }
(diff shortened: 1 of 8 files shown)


marcdumais-work:
Thanks for the review!

marcdumais-work:
Note: the dependency \[1\] that we added to the license-check ignore list in this PR, while it's being reviewed by the Eclipse Foundation IP team, is one used only for tests and not at runtime; i.e. it;s not distributed as part of the extension built from the code in this repository. So, I think we can safely go ahead and publish the extension without waiting for the IP review to be concluded, without fear that the dependency might end-up being deemed license-incompatible.

\[1\]: `npm/npmjs/-/playwright-core/1.46.1`

PR #273, opened 2024-09-11
Add docker example

frallax:
Add an example showing how to create a docker image that contains a Theia-based application with the vscode-trace-extension.

### DIFF
--- a/examples/docker/Dockerfile
+++ b/examples/docker/Dockerfile
@@ -1,0 +1,43 @@
+FROM node:18.19.1-bookworm-slim as build
+
+RUN apt-get update && apt-get install -y \
+    git \
+    python3 \
+    make \
+    pkg-config \
+    libsecret-1-dev \
+    g++ \
+    libx11-dev \
+    libxkbfile-dev
+
+# Build the browser trace-extension application
+COPY example-package.json /app/tte/package.json
+COPY webpack.config.js /app/tte/webpack.config.js
+WORKDIR /app/tte/
+RUN yarn && \
+    yarn run download:plugins && \
+    npx theia build --app-target=\"browser\" --mode production && \
+    yarn autoclean --init && \
+    echo *.ts >> .yarnclean && \
+    echo *.ts.map >> .yarnclean && \
+    echo *.spec.* >> .yarnclean && \
+    echo "!@theia/test" >> .yarnclean && \
+    yarn --production && \
+    yarn cache clean && \
+    du -hs /app/tte
+
+FROM node:18.19.1-bookworm-slim
+
+COPY --from=build /app/tte /app/tte
+
+RUN apt-get update && apt-get install -y \
+    libx11-6 \
+    libxkbfile1 \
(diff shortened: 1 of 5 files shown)


marcdumais-work:
I'm not sure if this contribution might fit-in better in repo `theia-trace-extension`? It's quite similar to the Docker example we already have in there, except this one here uses the `Trace Viewer for VSCode` from open-vsx, instead of using the `Theia Trace Viewer extension` from npm. 

Alternatively, since neither example application directly depend on the content in either repository, we might consider creating a new repo for them both, e.g. "Trace Viewer example applications" ?

@frallax  @bhufmann WDYT?

P.S. one potential argument to have this contribution in this repo here: we would need to try to confirm, but it's possible that the repo's tests, currently used against openvscode server running the vscode trace viewer, could be easily re-used on the containerized example Theia app running the same.

frallax:
@marcdumais-work not sure if this can be used for the tests you mention above, mainly because the Theia application is not picking the "latest vscode-trace-extension master" but one of the released prebuilt vsix packages.

I think it belongs to this repo just to provide an example of how the artifacts produced by this repo can be used.
But this is just my opinion.

marcdumais-work:
> not sure if this can be used for the tests you mention above, mainly because the Theia application is not picking the "latest vscode-trace-extension master" but one of the released prebuilt vsix packages.

I agree - but the goal would not be to test the trace viewer extension (it's already being tested), but rather to test the example application itself, to detect if it breaks. It's just a little bonus, if we can do that re-using the existing tests in the repo, like we did for the example application in theia-trace-extension.

bhufmann:
> I'm not sure if this contribution might fit-in better in repo `theia-trace-extension`? It's quite similar to the Docker example we already have in there, except this one here uses the `Trace Viewer for VSCode` extension from open-vsx, instead of using the `Theia Trace Viewer extension` from npm.
> 
> Alternatively, since neither example application directly depend on the content in either repository, we might consider creating a new repo for them both, e.g. "Trace Viewer example applications" ?
> 
> @frallax @bhufmann WDYT?
> 
> P.S. one potential argument to have this contribution in this repo here: we would need to try to confirm, but it's possible that the repo's tests, currently used against openvscode server running the vscode trace viewer, could be easily re-used on the containerized example Theia app running the same.

I think that this docker example should be in the `theia-trace-extension` repository, since it's a Theia application. @frallax could you please contribute it there?

frallax:
@bhufmann before following your advice, I'd like to understand the rationale of having this example in the theia-trace-extension repo. What is proposed in this pull request is an example of how to use the artifact created by this repo (vscode-trace-extension). That's why I contributed it to this repo.

The fact that the artifact is wrapped in a Theia application does not sound as a reason to have it in the theia-trace-extension repo. Maybe I should change it to a VSCode application, but then, would that really make a difference? Aren't we just changing the way we wrap an artifact created by the vscode-trace-extension repo?

Thanks for the explanation and help in advance!

PR #274, opened 2024-09-13
Select experiment when opening from open command

### LINKED ISSUE #264: Views view not refreshed when opening a new trace the first time
Opening a trace either using the file explorer or the open trace button, doesn't refresh the `VIEWS` view after the trace viewer tab with widgets appear and the new trace is not selected in the `OPENED TRACES` view.

vscode-no-views-refresh-at-open-trace

This change was introduced with pull request #250.

bhufmann:
Send TRACEVIEWERTAB_ACTIVATED signal when setting the experiment after a experiment is opened. This will select the experiment in the opened traces view which then will broadcast the EXPERIMENT_SELECTED signal to update other webviews.

Fixes #264

fix-selection-after-open


marcdumais-work (approved):
LGTM - small, elegant fix - Confirmed this fixes the issue without re-introducing the original endless switching problem

PR #276, opened 2024-10-21
Implement source lookup in Item Properties view

GregSavin:
The intention of this patch is to achieve the following runtime behavior:

If a loaded trace has a "Source" column, and if the value of that column is of the form file:line, then highlighting the event in question, and clicking on the "Source" value that shows up in Item Properties shall open the corresponding file in a VSCode editor, and position the cursor on the corresponding line number, if said file exists.

### DIFF
--- a/vscode-trace-common/src/messages/vscode-message-manager.ts
+++ b/vscode-trace-common/src/messages/vscode-message-manager.ts
@@ -74,7 +74,7 @@ export const VSCODE_MESSAGES = {
     OUTPUT_DATA_CHANGED: 'outputDataChanged',
     CONTRIBUTE_CONTEXT_MENU: 'contributeContextMenu',
     CONTEXT_MENU_ITEM_CLICKED: 'contextMenuItemClicked',
-    GO_TO_SOURCE_FILE: 'goToSourceFile'
+    SOURCE_LOOKUP: 'sourceLookup'
 };
 
 export class VsCodeMessageManager extends Messages.MessageManager {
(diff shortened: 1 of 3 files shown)


bhufmann:
@GregSavin thanks for the contribution. I missed this PR. I'll review it this week.

vscode-trace-common/src/messages/vscode-message-manager.ts
```diff
@@ -209,4 +210,9 @@ export class VsCodeMessageManager extends Messages.MessageManager {
         const data = JSON.stringify(payload);
         vscode.postMessage({ command: VSCODE_MESSAGES.CONTEXT_MENU_ITEM_CLICKED, data: data });
     }
+
+    sourceLookup(path: string, line: number): void {
+        const data = { path: path, line: line };
```
bhufmann, line 215: I troubleshooting a use case with the Trace Compass server and I noticed that the events table data provider for certain trace types will return a list of source in the form [file1:line1, file2:line2]. Even if it's only one file it will provide it with the brackets. I think when we did this prototype, there was only never the brackets. Anyways, I think for now it's ok to not support the list. I'll need to think about more how to support that better.

Is your format always `file:line` or is it also [file:line]`, i.e. with brackets?

vscode-trace-webviews/src/trace-explorer/properties/vscode-trace-explorer-properties-widget.tsx
```diff
@@ -68,7 +68,7 @@ class TraceExplorerProperties extends React.Component<{}, PropertiesViewState> {
             `${e.currentTarget.getAttribute('data-id')}`
         );
         console.log('filename: ' + fileLocation + ':' + line);
```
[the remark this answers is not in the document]
GregSavin: Thanks, I added the import for the CSS, and removed the console.log().

MatthewKhouzam (approved):
I like this, please address Bernd's comments first and consider this +1 the second review, I locally tested with codium

vscode-trace-common/src/messages/vscode-message-manager.ts
GregSavin: Thanks, I think I've addressed the immediately actionable review comments.   The fact that callstacks can carry source annotations (in a list form) is new to me.  Going forward, I'll try to learn more about that specific case (how the data flow works for that), but for now, our only usage of "Source" attribute is as an aspect of our ITmfEvent subclass, and it's only file:line.

PR #297, opened 2025-01-15
Update 'noExperiments' context after trace open

### LINKED ISSUE #296: Opening a Trace does not render UI properly if no traces are previously loaded on the server.
### Bug Description:
If the server is offline and you open a trace, and there are currently no existing traces stored on the server, the left-panel UI does not render correctly.
[Screencast from 01-15-2025 12:16:11 PM.webm](https://github.com/user-attachments/assets/cdea0c44-77ad-45bf-bc79-1c93aff6d46b)

### Steps to Reproduce:

1.  Either:
  a.  `rm -rf ./tracecompass-webapp`
  b. Start the server through `vscode-trace-extension`, close all opened experiments + traces, then stop the server with `ctrl + shift + p` => `Trace Server - stop or reset`
2.  From the views welcome UI on left panel (as shown above), and with the server offline, open a trace.  

The left panel displays that there are no opened traces, even though you just opened one.

### Additional Information
In this situation it seems that the backend is calling `updateNoExperimetnsContext` when the server is started, but not after the trace is opened.

The status of this `noExperiments` context determines if we render the welcome view or the trace-viewer left panel.

I'm adding a patch to update the `noExperiments` context after a trace / experiment is opened.

- Operating System: Ubuntu 22.04.5 LTS
- VSCode Version: 1.95.3

williamsyang-work:
### What it does
Fixes #296.
Triggers an update to the 'trace-explorer.noExperiments' context after a trace or experiment is opened.

### How to test
The steps to reproduce the bug can be found in #296.
The bug can be reproduced on `master`
The bug should be fixed on this branch.

### Follow-ups

When the server is online, the context is properly updated by a vscode-message sent from the frontend recieved by the backend.

#### Frontend:
```typescript
// 105-112 vscode-trace-explorer-opened-traces-widget.tsx
private initialized = false;
protected doHandleOpenedTracesChanged(payload: OpenedTracesUpdatedSignalPayload): void {
    if (!this.initialized) {
        this.initialized = true;
        return;
    }
    this._signalHandler.updateOpenedTraces(payload.getNumberOfOpenedTraces());
}
```
#### Backend:
```typescript
// 116-118 trace-explorer-opened-traces-webview-provider.ts
case VSCODE_MESSAGES.OPENED_TRACES_UPDATED:
    updateNoExperimentsContext();
    return;
```

I'm not sure why this code is triggered when the server is online and not when the server is offline.  I didn't dig too deep into investigating that because I'm not sure the relevancy, and I know that the events getting bounced around on the FE / BE can get really messy and hard to trace.  

There is also the tracking of `initialized` and we may simply not be sending the first signal intentionally.  This can be investigated because it may not be necessary to track anymore and could lead a cleaner solution.

I didn't deep dive into this.  I think updating the context on the backend after a trace is opened is a reasonable approach. WDYT?

### Review checklist

- [x ] As an author, I have thoroughly tested my changes and carefully followed the instructions in this template

### DIFF
--- a/vscode-trace-extension/src/trace-explorer/trace-utils.ts
+++ b/vscode-trace-extension/src/trace-explorer/trace-utils.ts
@@ -4,6 +4,7 @@ import { Trace as TspTrace } from 'tsp-typescript-client/lib/models/trace';
 import { TraceViewerPanel } from '../trace-viewer-panel/trace-viewer-webview-panel';
 import { getExperimentManager, getTraceManager } from '../utils/backend-tsp-client-provider';
 import { traceLogger } from '../extension';
+import { updateNoExperimentsContext } from '../utils/backend-tsp-client-provider';
 import { KeyboardShortcutsPanel } from '../trace-viewer-panel/keyboard-shortcuts-panel';
 
 // eslint-disable-next-line no-shadow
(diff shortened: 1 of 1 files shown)


vscode-trace-extension/src/trace-explorer/trace-utils.ts
```diff
@@ -62,93 +63,101 @@ export const fileHandler =
...
+                        panel.dispose();
+                        return;
                     }
-                    rollbackTraces(traces, 20, progress);
-                    progress.report({ message: ProgressMessages.COMPLETE, increment: 10 });
-                    panel.dispose();
-                    return;
+                    progress.report({ message: ProgressMessages.COMPLETE, increment: 30 });
+                } finally {
+                    updateNoExperimentsContext();
```
marcdumais-work, line 159: That's a creative way of ensuring to update that context after that "leaky" block executes, without adding it in a bunch of places - I like

marcdumais-work (approved):
LGTM - thanks for this contribution!

marcdumais-work:
> I think updating the context on the backend after a trace is opened is a reasonable approach. WDYT?

seems reasonable to me

bhufmann (approved):
Looks good to me. This change fixed it. Thanks!

PR #300, opened 2025-01-28
Add extra configuration parameter for vscode backend URL

frallax:
### What it does

The vscode-trace-extension code is written so that some REST calls towards the server are done by the vscode backend, and some from the vscode frontend (webviews). In some situations can be useful to be able to specify different URLs for the REST calls, depending on where the REST call is done (vscode frontend or backend). For example when deploying on a k8s cluster, where the vscode backend can communicate to the trace server without going through an ingress.

This PR add the setting to be able to select the URL used from the vscode backend. 

### How to test

To make sure that the added feature works as expected (i.e. the REST calls are done to different URLs and/or ports), a simple test setup can be configured locally by:

- starting 2 trace-servers on 2 different ports
  - ```trace-server -data ~/.trace-server-webapp8080 -vmargs -Dtraceserver.port=8080 -Dtraceserver.host=127.0.0.1```
  - ```trace-server -data ~/.trace-server-webapp8081 -vmargs -Dtraceserver.port=8081 -Dtraceserver.host=127.0.0.1```

- apply the following patch to the tsp-typescript-client project (dependency of vscode-trace-ext) and use it to build the vscode-trace-ext from this branch (see yarn link instructions on the project main readme)

```
-- a/tsp-typescript-client/src/protocol/rest-client.ts
+++ b/tsp-typescript-client/src/protocol/rest-client.ts
@@ -104,6 +104,7 @@ export class RestClient {
         body?: any,
         normalizer?: Normalizer<T>,
     ): Promise<TspClientResponse<Deserialized<T>>> {
+        console.log(`[${this.feOrBe()}] TSP_client#performRequest() - method: ${method}, url: ${url}, body: ${this.jsonStringify(body)} `);
         let response: HttpResponse;
         try {
             response = await this.httpRequest({
@@ -167,6 +168,16 @@ export class RestClient {
         return JSONBig.stringify(data);
     }
 
+    protected static feOrBe(): string {
+        const isBrowser = typeof window !== "undefined" && window instanceof Window;
+
+        if (isBrowser) {
+            return "FRONTEND"
+        }
+        
+        return "BACKEND"
+    }
+
     /**
      * Parse JSON-encoded data. If a number is too large to fit into a regular
      * `number` then it will be deserialized as `BigInt`.
diff --git a/tsp-typescript-client/tsconfig.json b/tsp-typescript-client/tsconfig.json
index 5f39976..4509bc5 100644
--- a/tsp-typescript-client/tsconfig.json
+++ b/tsp-typescript-client/tsconfig.json
@@ -7,7 +7,7 @@
         "module": "commonjs",
         "moduleResolution": "node",
         "lib": [
-            "es6"
+            "es6","DOM", "ES2020"
         ],
         "alwaysStrict": true,
         "strictNullChecks": true,
 ```

- Run the vscode-trace-extension in development mode, access the settings and enable the backend URL setting it to localhost:8081

- Try to open a trace and look into the console (browser devtools). It will show which calls are done from the frontend/backend and where are they going. Of course should be going to the URLs specified in the settings :)

### Follow-ups

No follow ups planned as of now

### Review checklist

- [x ] As an author, I have thoroughly tested my changes and carefully followed the instructions in this template

### DIFF
--- a/vscode-trace-extension/package.json
+++ b/vscode-trace-extension/package.json
@@ -258,12 +258,26 @@
                 "trace-compass.traceserver.url": {
                     "type": "string",
                     "default": "http://localhost:8080",
-                    "description": "Enter the trace server's URL, including port. Eg: http://localhost:8080."
+                    "description": "Enter the trace server's URL, including port. Eg: http://localhost:8080.",
+                    "order": 0
                 },
                 "trace-compass.traceserver.apiPath": {
                     "type": "string",
                     "default": "tsp/api",
-                    "description": "Enter the trace server's API path, to be appended to the server URL. Eg: 'tsp/api'."
+                    "description": "Enter the trace server's API path, to be appended to the server URL. Eg: 'tsp/api'.",
+                    "order": 1
+                },
+                "trace-compass.traceserver.enableSeparateBackendUrl": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Enable separate backend URL - Allows the vscode-backend to use a different URL. This is useful for a remote-deployment with K8s where the frontend needs to access an ingress, while backend needs to access the same trace-server using an internal URL.",
+                    "order": 2
+                },
+                "trace-compass.traceserver.backendUrl": {
+                    "type": "string",
+                    "default": "http://localhost:8080",
+                    "description": "Backend trace-server URL - Only used when separate backend URL is enabled.",
+                    "order": 4
                 }
             }
         },
--- a/vscode-trace-extension/src/utils/backend-tsp-client-provider.ts
+++ b/vscode-trace-extension/src/utils/backend-tsp-client-provider.ts
@@ -9,32 +9,41 @@ import { TspClientProvider } from 'vscode-trace-common/lib/client/tsp-client-pro
  *  for the Trace Extension's VSCode Backend
  */
 
-let _root: string;
-let _path: string;
-let _url: string;
+let _rootBE: string;
+let _rootFE: string;
+let _urlBE: string;
+let _urlFE: string;
 let _provider: TspClientProvider;
 
-export const getTraceServerUrl = (): string => _root;
-export const getTspApiEndpoint = (): string => _path;
-export const getTspClientUrl = (): string => _url;
+export enum ClientType {
+    BACKEND = 'backend',
+    FRONTEND = 'frontend'
+}
+
+export const getTraceServerUrl = (clientType: ClientType): string =>
+    clientType === ClientType.FRONTEND ? _rootFE : _rootBE;
+export const getTspClientUrl = (clientType: ClientType): string =>
+    clientType === ClientType.FRONTEND ? _urlFE : _urlBE;
 
 export const getTspClient = (): TspClient => _provider.getTspClient();
 export const getExperimentManager = (): ExperimentManager => _provider.getExperimentManager();
 export const getTraceManager = (): TraceManager => _provider.getTraceManager();
 
-export const updateTspClientUrl = async (): Promise<string> => {
-    const _extUri = await getExternalUriFromUserSettings();
-    _root = _extUri.toString();
-    _path = getApiPathFromUserSettings();
-    _url = _root + _path;
+export const updateTspClientUrl = async (): Promise<void> => {
+    const extUri = await getExternalUriFromUserSettings(ClientType.BACKEND);
+    _rootBE = extUri.toString();
+    const apiPath = getApiPathFromUserSettings();
+    _urlBE = _rootBE + apiPath;
+
+    const extUriFE = await getExternalUriFromUserSettings(ClientType.FRONTEND);
+    _rootFE = extUriFE.toString();
+    _urlFE = _rootFE + apiPath;
 
     if (!_provider) {
-        _provider = new TspClientProvider(_url, undefined);
+        _provider = new TspClientProvider(_urlBE, undefined);
     } else {
-        _provider.updateTspClientUrl(_url);
+        _provider.updateTspClientUrl(_urlBE);
     }
-
-    return _url;
 };
 
 export const addTspClientChangeListener = (listenerFunction: (tspClient: TspClient) => void): void => {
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

**Comment by marcdumais-work** [comment]
I believe the license check failure is unrelated to this PR, so please ignore it for now.

File: vscode-trace-extension/src/utils/backend-tsp-client-provider.ts [hunk 1]
```diff
@@ -9,32 +9,36 @@ import { TspClientProvider } from 'vscode-trace-common/lib/client/tsp-client-pro
...
     if (!_provider) {
-        _provider = new TspClientProvider(_url, undefined);
+        _provider = new TspClientProvider(_urlBE, undefined);
     } else {
-        _provider.updateTspClientUrl(_url);
+        _provider.updateTspClientUrl(_urlBE);
     }
 
-    return _url;
+    return _urlBE;
```
bhufmann, line 41: Shouldn't `_urlFE` be returned here. This is called to update the front-end. Maybe to make it clearer is to not have this method return anything, but call getTraceServerUrl(true) to get the URL to propagate. See extension.ts code above where it is propagated when the URL is changed.

bhufmann (changes requested):
I tested it with the code printout to the console from the tsp-typescript-client and I can confirm that the BE url is used when TSP calls are done from the BE (health check and during open procedures). If disabled, it works as before.

Maybe add a chapter to the README about this setup. 

When using [Remote SSH](https://github.com/eclipse-cdt-cloud/vscode-trace-extension?tab=readme-ov-file#remote-ssh-support), then the split setting needs to be disabled because the port can only be forwarded once. I don't think there needs to be any code changes for that (unless you can think about something). However, the corresponding [README chapter](https://github.com/eclipse-cdt-cloud/vscode-trace-extension?tab=readme-ov-file#remote-ssh-support) should have a note about it.

vscode-trace-extension/src/utils/backend-tsp-client-provider.ts
frallax: Opted to remove the return value and use ```etTspClientUrl(true);```
If you agree please resolve the conversation, or let me know if I should do adjustments.

The README has also been updated.

bhufmann (changes requested):
Thanks for the updates. I re-tested and reviewed it. It looks good to me. I'd like also @marcdumais-work to have a look because he's been part of the discussion.

The linting check still fails. You can fix it by running `yarn format:write`.

Could you please squash the commits and write a commit message for the change. Add both your and @williamsyang-work signed-off to the commit message since he contributed to this change as well.

vscode-trace-extension/package.json
```diff
@@ -258,12 +258,26 @@
...
                     "type": "string",
                     "default": "tsp/api",
-                    "description": "Enter the trace server's API path, to be appended to the server URL. Eg: 'tsp/api'."
+                    "description": "Enter the trace server's API path, to be appended to the server URL. Eg: 'tsp/api'.",
+                    "order": 1
+                },
+                "trace-compass.traceserver.enableSeparateBackendUrl": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Enable separate backend URL - Allows the vscode-backend to use a different URL. This is useful for a remote-deployment with K8s where the frontend needs to access an ingress, while backend needs to access a localhost for trace-server on the same machine.",
```
marcdumais-work, line 273: I would suggest making it explicit that in the "separate BE URL" mode, that URL is meant to point to the same trace server, not a second one. Maybe a slight rewording would do the trick: I was not sure about the part "on the same machine" - it could be on a different machine I think? If so, what about something like: "[...] while backend needs to access the same trace-server using an internal URL"? Or maybe there would be a better way to express similar?

vscode-trace-extension/src/utils/backend-tsp-client-provider.ts
```diff
@@ -67,9 +69,16 @@ export async function updateNoExperimentsContext(): Promise<void> {
...
 
-async function getExternalUriFromUserSettings(): Promise<vscode.Uri> {
+async function getExternalUriFromUserSettings(isFrontEnd: boolean): Promise<vscode.Uri> {
     const tsConfig = vscode.workspace.getConfiguration('trace-compass.traceserver');
-    const traceServerUrl: string = tsConfig.get<string>('url') || 'http://localhost:8080';
+    let traceServerUrl: string;
+    if (isFrontEnd) {
+        traceServerUrl = tsConfig.get<string>('url') || 'http://localhost:8080';
+    } else {
+        traceServerUrl = tsConfig.get<string>('enableSeparateBackendUrl')
```
marcdumais-work, line 78: should this be "<boolean>" instead of "<string>"?

vscode-trace-extension/src/utils/backend-tsp-client-provider.ts
```diff
@@ -67,9 +69,16 @@ export async function updateNoExperimentsContext(): Promise<void> {
     return;
 }
 
-async function getExternalUriFromUserSettings(): Promise<vscode.Uri> {
+async function getExternalUriFromUserSettings(isFrontEnd: boolean): Promise<vscode.Uri> {
```
marcdumais-work, line 72: Question: have you considered using an Enum instead of a Boolean? It might make the code easier to understand, specially from the calling code. 

Same could apply for getTraceServerUrl() and  getTspClientUrl()

marcdumais-work:
Looks good generally - I have added a couple of minor comment plus a question.

marcdumais-work:
Note: there is currently one dependency failing the automated license check. We have the same on the master branch - that dependency used to pass the check but suddenly started to fail. Since the failure is unrelated to this PR, I think we can ignore it here. 

For the record, there is an IP ticket for the dependency in question:
[main] INFO A review is required for npm/npmjs/@types/parse-json/4.0.2.
[main] INFO A review request already exists https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/19199

frallax:
> Thanks for the updates. I re-tested and reviewed it. It looks good to me. I'd like also @marcdumais-work to have a look because he's been part of the discussion.
> 
> The linting check still fails. You can fix it by running `yarn format:write`.
> 
> Could you please squash the commits and write a commit message for the change. Add both your and @williamsyang-work signed-off to the commit message since he contributed to this change as well.

Sorry for the delay but I have been sick the last few days.
I have fixed the linting and addresses @marcdumais-work comments. Once marc gives me his blessing I will squash.

vscode-trace-extension/src/utils/backend-tsp-client-provider.ts
frallax: I tried applying a enum. Let me know if you like it better now.

vscode-trace-extension/src/utils/backend-tsp-client-provider.ts
frallax: Fixed

vscode-trace-extension/package.json
marcdumais-work: Looks good, thanks!

vscode-trace-extension/src/utils/backend-tsp-client-provider.ts
frallax: Done. The branch is also squashed, so if there are no other concerns/comments it should be ready to go

marcdumais-work (approved):
LGTM - thanks @frallax  and @williamsyang-work !

bhufmann (approved):
Looks good to me. Thanks!

PR #314, opened 2025-03-25
[DO NOT MERGE] PoC: Traceviewer packages as git subtree

marcdumais-work:
### What it does

This is a PoC, experimenting bringing the two `trace-viewer-*` packages in source form locally in this repo, while maintaining them as-is in their original repo (theia-trace-extension). 

In this PoC, repo `theia-trace-extension` was added under directory `packages` in "squashed form" (keeping the bulk of the source code but no the whole history, which is still available in the original repo).
 
```bash
# add remote:
$ git remote add -f traceviewer-packages git@github.com:eclipse-cdt-cloud/theia-trace-extension.git
Note: fetches history including tags - show we filter these out? Or maybe just delete them locally?

# fetch subtree:
$ git subtree add --prefix packages traceviewer-packages master --squash
```

Then the 2 packages of interest were added as elements of the root `package.json` workspace and a few minor changes were made to accommodate the 2 new components building, outside of their usual context.

In theory, with this setup, one can make changes that cover both the usual components of this repo as well as `traceviewer-base` and/or `traceviewer-react-components`, as if the `traceviewer-*` components were local -  they are, in a way. It may be easier later to upstream, if changes to the subtree are kept in separate commits vs changes in the rest of the repo (they share git history in this setup). 

Then, when wanted, one could push the git subtree changes toward the upstream (theia-trace-extensions) as a PR and then merge and eventually publish the components to npm from there.

Questions / limitations:

versioning:
How to handle versioning of the `traceviewer-*` packages? Since the packages are part of the `package.json` workspace, our lerna publish commands will step their versions too, but we want to manage these versions in their original repo, not here where they are essentially used as private packages.

Changes in the subtree vs upstream:
The subtree contains the whole `theia-trace-extension` repository (that's just how a subtree works, AFAIK), but only the two `traceviewer-*` components are "integrated", by having their folders added in the `package.json` workspaces list. I expect that pure code changes in these components will be pretty solid, but dependencies changes could have to be tweaked when upstreaming to `theia-trace-extension` - in the local subtree setup, we are effectively by-passing the other workspaces present in the upstream repo and also the upstream `yarn.lock`,  which could cause discrepancies.

e.g.: I expect it would not be a good idea to attempt to run `yarn upgrade` from the subtree, hoping to then upstream the result to `theia-trace-extension` repo. 

Maintenance of the subtree
Once in a while it will probably be necessary to update the subtree to reflect upstream changes, even if ATM most of the code changes happen in the subtree. e.g. after running `yarn upgrade` upstream and adjusting some dependencies' ranges. 

### How to test
 

It would be good to try some small changes that cover the subtree and the extension, and attempt to push the subtree changes as a PR top theia-trace-extension, to confirm how it works.

Also good to test updating the subtree after `yarn upgrade` upstream.

### Review checklist

- [ ] As an author, I have thoroughly tested my changes and carefully followed the instructions in this template


marcdumais-work:
No longer relevant - closing.

Issue #307, opened 2025-03-06
Pssible event listener leak in webviews, when starting without a trace server


marcdumais-work:
### Bug Description:

When testing something else, I noticed the following in the debug console:

```
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 EXPERIMENT_SELECTED listeners added to [SignalManager]. MaxListeners is 10. Use emitter.setMaxListeners() to increase limit
(Use `exe --trace-warnings ...` to show where the warning was created)
```

I used the debugger to confirm - it looks like when the extension is started without a trace server, the webviews are created and at least 4 of them registers a listener for event `EXPERIMENT_SELECTED` (some views register other event listeners too, but that one is the first to hit the "max 10" limit). 

Then when the trace server is started, it looks like the same 4 webviews are probably re-created or at least go through their init again, registering new listeners. 

From there it's just a natter of opening a few traces and we go over the 10 listeners limit for event `EXPERIMENT_SELECTED`. 

### Steps to Reproduce:

1.
2.
3.

### Additional Information

- Operating System: Ubuntu 22.04
- VSCode Version: VSCodium v1.97.2

marcdumais-work:
From what I can see, even without the leak (which adds 4 listeners), we would get the warning with ~7 traces open, which seems a not-unreasonable use-case. Maybe we should consider increasing the limit as per the warning.

marcdumais-work:
I think the issue may be related to this line, towards the end of the extension's `activate()`, where momentarily we simulate the trace server being up (when sometimes it's not yet):

![Image](https://github.com/user-attachments/assets/a891a9e1-52cf-4edf-86ea-aa0327bc3dab)

PR #315, opened 2025-03-25
Minimize jitter related to creation of trace explorer webview views

marcdumais-work:
### What it does

The initialization of webview views can be tricky. There are the 4 trace explorer views and also welcome view, one set of which should be displayed at any given moment, depending on context variables. However it can be possible to have jitter initially if the context variables change value quickly, which can result in webview views being resolved again without the "onDispose" event being sent. When this happens, the webview views have no opportunity to unregister their event listeners, causing leaks.

To address the jitter, context variables `trace-explorer.noExperiments` and `traceViewer.serverUp` are initialized at activation with values that will prevent the trace explorer webview views being "resolved", causing an unambiguous situation where the welcome view will be displayed instead, until a trace server is started or detected to be running, and through communication with it, it's established that one or more experiment exists on it.
                               
In case there might still be jitter in cases that were missed, or in case some in re-introduced in later changes, we catch the situation where a trace explorer webview view is re-resolved without having been disposed and in that case dispose of its listeners and print a warning message on the console.

Also, corrected that one of the trace explorer webview views, `TraceExplorerItemPropertiesProvider`, was not calling "super.dispose()" in its own dispose() method. Since this is where the "resolved" webview is set, this would trip the warning added in this commit, even if the old view had been disposed correctly.

### How to test

Scenarios: 

Try all possible combinations of starting the trace viewer extension with and without a trace server and an experiment already present or not, and confirm that no webview view leaks event listeners (i.e. that they are all correctly disposed). Also confirm that all scenarios involving further starting/stopping the trace server and creating/deleting an experiment, still works as before. 

Fresh start with and without a trace server already started, with and without an experiment already present. 

There is a temporary commit present, that adds some debug traces to help test, that will be removed before merging. Using that commit, it's possible to confirm whether the trace explorer's webview views are disposed and as well follow important changes in value for the context variables. 

For example, if the trace server is already started and has an experiment, the following debug traces will be printed at startup:

_isTraceServerUp() - set traceviewer.serverUp: true
extensionHostProcess.js:178
updateServerStatus() - set traceviewer.serverUp: true
extensionHostProcess.js:178
isTraceServerUp() - set traceviewer.serverUp: true
extensionHostProcess.js:178
updateNoExperimentsContext - set trace-explorer.noExperiments: false
extensionHostProcess.js:178_
__vscode calling resolveWebviewView() to initialize webview: TraceExplorerOpenedTracesViewProvider
extensionHostProcess.js:178
vscode calling resolveWebviewView() to initialize webview: TraceExplorerAvailableViewsProvider
extensionHostProcess.js:178
vscode calling resolveWebviewView() to initialize webview: TraceExplorerTimeRangeDataProvider
extensionHostProcess.js:178
vscode calling resolveWebviewView() to initialize webview: TraceExplorerItemPropertiesProvider
extensionHostProcess.js:178__

Then, closing the experiment, we can see the 4 webview views being disposed:

_updateNoExperimentsContext - set trace-explorer.noExperiments: true
extensionHostProcess.js:178_
__Cleaning-up after vscode disposed webview: TraceExplorerItemPropertiesProvider
extensionHostProcess.js:178
Cleaning-up after vscode disposed webview: TraceExplorerTimeRangeDataProvider
extensionHostProcess.js:178
Cleaning-up after vscode disposed webview: TraceExplorerAvailableViewsProvider
extensionHostProcess.js:178
Cleaning-up after vscode disposed webview: TraceExplorerOpenedTracesViewProvider__

It's also possible to use the debugger to confirm whether there are still event listener leaks - see #307 for the details how to set that up.

### Follow-ups

N/A
### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed the instructions in this template

### DIFF
--- a/vscode-trace-extension/src/trace-explorer/abstract-trace-explorer-provider.ts
+++ b/vscode-trace-extension/src/trace-explorer/abstract-trace-explorer-provider.ts
@@ -60,6 +60,20 @@ export abstract class AbstractTraceExplorerProvider implements vscode.WebviewVie
         _context: vscode.WebviewViewResolveContext,
         _token: vscode.CancellationToken
     ): void {
+        if (this._view) {
+            // The initialization of webview views can be tricky. There are the 4 trace
+            // explorer views and also welcome view, one set of which should be displayed
+            // at any given moment, depending on context variables. However it can be
+            // possible to have jitter initally if the context variables change value
+            // quickly, which can result in webview views being resolved again without
+            // the "onDispose" event being sent. We can detect and correct for this
+            // condition, in case a change in the code brings it back again
+            console.warn(
+                'AbstractTraceExplorerProvider#resolveWebviewView(): unexpectedly called again without first disposing of old webview view: ' +
+                    this.constructor.name
+            );
+            this.dispose();
+        }
         this._view = webviewView;
         webviewView.webview.html = this._getHtmlForWebview(webviewView.webview);
 
--- a/vscode-trace-extension/src/extension.ts
+++ b/vscode-trace-extension/src/extension.ts
@@ -39,6 +39,10 @@ export const messenger = new Messenger({ ignoreHiddenViews: false });
 export async function activate(context: vscode.ExtensionContext): Promise<ExternalAPI> {
     traceLogger = new TraceExtensionLogger('Trace Extension');
 
+    // Initialize noExperiments/serverUp in a way to avoid WebviewViews creation jitter
+    vscode.commands.executeCommand('setContext', 'trace-explorer.noExperiments', true);
+    vscode.commands.executeCommand('setContext', 'traceViewer.serverUp', false);
+
     const resourceTypeHandler: TraceExplorerResourceTypeHandler = TraceExplorerResourceTypeHandler.getInstance();
 
     await updateTspClientUrl();
@@ -239,10 +243,6 @@ export async function activate(context: vscode.ExtensionContext): Promise<Extern
 
     vscode.commands.executeCommand('setContext', 'traceViewer.markerSetsPresent', false);
     vscode.commands.executeCommand('setContext', 'traceViewer.markerCategoriesPresent', false);
-
-    // Initialize noExperiments/serverUp in a way so that trace explorer webviews are initialized
-    vscode.commands.executeCommand('setContext', 'trace-explorer.noExperiments', false);
-    vscode.commands.executeCommand('setContext', 'traceViewer.serverUp', true);
 
     // Refresh to trigger rendering trace explorer or welcome page
     vscode.commands.executeCommand('trace-explorer.refreshContext');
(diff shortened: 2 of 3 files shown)


vscode-trace-extension/src/trace-explorer/abstract-trace-explorer-provider.ts
```diff
@@ -60,6 +60,10 @@ export abstract class AbstractTraceExplorerProvider implements vscode.WebviewVie
         _context: vscode.WebviewViewResolveContext,
         _token: vscode.CancellationToken
     ): void {
+        if (this._view) {
```
bhufmann, line 63: I wonder why the dispose() method is not called for the initial view instances. Any ideas?

marcdumais-work: It's a good question - I will dig to try to find out.

marcdumais-work: I could not find an exact cause, but I have a theory:  

At any given time, either the trace explorer sub views (webview views) or the welcome view can be visible, as determined by the value of two context variables: `trace-explorer.noExperiments` and `traceViewer.serverUp`, as defined in the `when` clause for each view:

(package.json):
image

We create one `[...]ViewsProvider`object for each trace explorer webview view and register them with vscode, then vscode evaluates whether and how they should be displayed, according to the `package.json` configs. When a webview view becomes visible, vscode creates it and passes it to us, calling its `resolveWebviewView()` method (ours is in `AbstractTraceExplorerProvider` and used for all our trace explorer webview views). When a webview view becomes not visible, vscode is supposed to discard it which should trigger `onDidDispose()` that we register in `resolveWebviewView()`. When the webview view becomes visible again, vscode re-creates it and calls `resolveWebviewView()` again, providing the new webview object (that has a different UUID).

From experimentation, I noticed that if I quickly changed the value for the two context variables during activation, multiple sets of listeners leaked (vs the one set of 4, observed before). 

Theory: vscode does not expect or does not perfectly handle quick jitter, that causes webviews to be quickly displayed and then hidden (created/disposed)), which was happening during our extension's activation. Perhaps hiding a webview view before it has finished initializing (on vscode side), causes it to not correctly be disposed?

When the context variables were set to minimize jitter (i.e. only show trace explorer when we really have detected the trace server is up and has at least one experiment), I could see that all webview views were then correctly disposed when hidden, triggering our `onDispose()` method and performing the expected cleanup of the previous instance's listeners.

bhufmann (commented):
It looks good to me. I tested the following cases:
- start of extension with external server running and experiments available: server is up and open traces view is populated
- start of extension with external server running with no experiments: server is up and welcome view is shown
- start and stop of trace server is handled correctly
- redo the tests above with vscode-remote setup
- switch between trace-server by changing preferences (e.g. URL): correct trace server is queried, server is up and opened traces view populated.

The following case is not working with and without this change, and it should be handled outside this PR.

- start of extension with external server running. Views are populated properly. Kill external server, select file explorer and then trace explorer. Neither trace-server status in status bar shows not connected nor the welcome view is shown.  

Please remove the print-outs and then I'll approve it.

bhufmann:
> * start of extension with external server running. Views are populated properly. Kill external server, select file explorer and then trace explorer. Neither trace-server status in status bar shows not connected nor the welcome view is shown.  I need to press refresh button to fix it.

Bug report for this issue: https://github.com/eclipse-cdt-cloud/vscode-trace-extension/issues/316

marcdumais-work:
> Please remove the print-outs and then I'll approve it.

Done - I removed the top commit containing the print-outs

marcdumais-work:
Thanks for the review!

PR #344, opened 2025-08-19
Add doc guide for developing a new view

PatrickTasse:
### What it does

Add doc guide for developing a new view

### How to test

N/A

### Follow-ups

N/A

### Review checklist

- [ ] As an author, I have thoroughly tested my changes and carefully followed the instructions in this template

### DIFF
--- a/doc/Develop-new-view.md
+++ b/doc/Develop-new-view.md
@@ -1,0 +1,183 @@
+# Developing a new view
+
+This guide will describe the mandatory and optional steps necessary to implement a new view for the VSCode Trace Viewer. It covers the full development stack, including the Trace Compass trace server, the Trace Server Protocol (TSP) and impacted libraries used by the VSCode Trace Viewer.
+
+## Data Provider
+
+If the new view is to be populated using a data model that is not already available from the trace server, then a new data provider will need to be implemented.
+
+The data provider's role is to provide a model corresponding to the requested method and parameters, possibly by scheduling an analysis and querying its resulting output, or any other method to extract data from the trace and compute the appropriate result.
+
+## Data Provider Factory / Descriptor
+
+To be visible to the VSCode Trace Viewer client, the new view should be declared using the org.eclipse.tracecompass.tmf.core.dataprovider extension point.
+
+```xml
+<extension point="org.eclipse.tracecompass.tmf.core.dataprovider">
+    <dataProviderFactory
+        class="org.foo.FooFactory"
+        id="org.foo.factory.id">
+    </dataProviderFactory>
+</extension>
+```
+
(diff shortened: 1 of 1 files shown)


doc/Develop-new-view.md
```diff
@@ -0,0 +1,183 @@
+# Developing a new view
+
+This guide will describe the mandatory and optional steps necessary to implement a new view for the VSCode Trace Viewer. It covers the full development stack, including the Trace Compass trace server, the Trace Server Protocol (TSP) and impacted libraries used by the VSCode Trace Viewer.
+
+## Data Provider
+
+If the new view is to be populated using a data model that is not already available from the trace server, then a new data provider will need to be implemented.
+
+The data provider's role is to provide a model corresponding to the requested method and parameters, possibly by scheduling an analysis and querying its resulting output, or any other method to extract data from the trace and compute the appropriate result.
```
marcdumais-work, line 9: quibble: add a comma after "or any other method" ?  Alternatively, this is a pretty long sentence that could be broken-down into 2.

doc/Develop-new-view.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Data Provider Factory / Descriptor
+
```
bhufmann, line 12: The guide below is specific for the Trace Compass server. I wonder if this should be part of the trace server specific developer guide: https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/blob/master/doc/trace-server/trace-server-dev-guide.md?

I recognize that the documentation is already scattered, for example, the referenced trace server guide above has specifics for Trace Compass core AND the trace server service layer which probably should be in the repo with the trace server code. Not sure how we should handle best full-stack development that affects different layers.

I also realized that it would be hard for someone else that doesn't use the Trace Compass server to propose changes to the TSP, because the TSP yaml is generated by the Trace Compass server.

Suggestions for improving the situation is welcome.

doc/Develop-new-view.md
```diff
@@ -0,0 +1,183 @@
...
+    @Parameter(description = "exp UUID") @PathParam("expUUID") UUID expUUID,
+    @Parameter(description = "output ID") @PathParam("outputId") String outputId,
+    @RequestBody(description = "Query parameters to fetch the Foo model",
+            content = { @Content(examples = @ExampleObject("{\"parameters\":{ \"key\": \"value\" }}"),
+            schema = @Schema(implementation = FooQueryParameters.class)) },
+            required = true)
+        QueryParameters queryParameters) {
+```
+
+## TSP Typescript Client
```
bhufmann, line 111: Here the FE changes are outlined for a new view type. First the protocol level.

doc/Develop-new-view.md
```diff
@@ -0,0 +1,183 @@
+# Developing a new view
```
bhufmann, line 1: should it be view type?

MatthewKhouzam:
Any updates?

PR #395, opened 2026-05-14
Integrate rich cursor and fix tooltip pointer events

MatthewKhouzam:
### What it does
- Add TimeGraphChartRichCursor layer to gantt chart
- Set tooltip pointerEvents to 'none' to avoid blocking canvas events
- Add pixi.js-legacy resolution to force 7.x from linked timeline-chart

### How to test

Open a timegraph

### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed the instructions in this template

## Summary by CodeRabbit

* **New Features**
  * Added a rich cursor visualization and interaction layer to the main chart, providing clearer pointer feedback during timeline exploration.

* **Bug Fixes**
  * Updated tooltips so they no longer capture pointer interactions, preventing tooltip hover behavior from interfering with other chart and interface actions.

* **Chores**
  * Updated chart rendering support to improve compatibility and consistency across supported environments.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     ],
     "resolutions": {
         "@vscode/vsce": "2.25.0",
-        "@types/react": "18.3.8"
+        "@types/react": "18.3.8",
+        "pixi.js-legacy": "^7.3.2"
     }
 }
--- a/local-libs/traceviewer-libs/react-components/package.json
+++ b/local-libs/traceviewer-libs/react-components/package.json
@@ -34,7 +34,7 @@
         "react-grid-layout": "1.2.0",
         "react-modal": "^3.8.1",
         "react-virtualized": "^9.21.0",
-        "timeline-chart": "^0.4.3",
+        "timeline-chart": "^0.5.0",
         "traceviewer-base": "0.11.0",
         "tsp-typescript-client": "^0.9.0"
     },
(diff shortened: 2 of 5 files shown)


MatthewKhouzam:
@jonatanantoni and @thorstendb-ARM

jreineckearm:
Requires https://github.com/eclipse-cdt-cloud/timeline-chart/pull/338

JonatanAntoni:
I confirm the cursor on a time-line chart gets a row-based overlay label reflecting the full state label at the cursor position.

I do observe a strange behaviour though: Initially, after clicking into the time chart canvas, the cursor moves along with the mouse. But after a while the cursor becomes sticky and does *not* follow the mouse movement anymore. Once I click again into the canvas, the cursor moves again.

Opinion: I'd expect the cursor to span all visible time-based graphs and being shown in a similar way. With this change, the cursor on other views is just set the the time position where the mouse click appears but does *not* move along with the mouse pointer, nor does it show similar signal markers. E.g., in a time XY with multiple signals, we should have current value displayed at the cursor position as well while the cursor follows the mouse pointer.

package.json
```diff
@@ -45,6 +45,7 @@
     ],
     "resolutions": {
         "@vscode/vsce": "2.25.0",
-        "@types/react": "18.3.8"
+        "@types/react": "18.3.8",
+        "pixi.js-legacy": "^7.3.2"
```
JonatanAntoni, line 49: Do we need to repeat this transitive dependency here?

MatthewKhouzam: This one is a big one, we upgraded from 5 - 7, but maybe no, we can test later! ;)

MatthewKhouzam: It is needed! I tried without and it failed. Maybe it's a matthew problem though.

MatthewKhouzam:
@arfio any thoughts as to why the uplift didn't work?

PR #404, opened 2026-07-09
Add legends column to XY trace viewer

hoathaitran3209:
## Summary

This pull request adds a legends column to the XY trace viewer to improve readability and make it easier to identify plotted series.

## Changes

- Add a legends column to the filter tree.
- Update XY output components to support the new column.
- Adjust table layout and styles.
- Refactor related utility code where necessary.

## Testing

- Verified the legends column is displayed correctly.
- Verified existing XY trace viewer functionality continues to work as expected.

## Summary by CodeRabbit

* **New Features**
  * Added per-series legend color swatches to XY charts and the filter tree/table.
  * Introduced a dedicated **“Legend”** column to keep legend labeling consistent across views.
* **Bug Fixes**
  * Legend colors now initialize and refresh correctly whenever XY data is rebuilt and when tree/table content updates.
  * Updated rendering so legend/color changes trigger the correct re-renders.
* **Style**
  * Added styling for compact legend swatches.

### DIFF
--- a/local-libs/traceviewer-libs/react-components/src/components/abstract-xy-output-component.tsx
+++ b/local-libs/traceviewer-libs/react-components/src/components/abstract-xy-output-component.tsx
@@ -23,10 +23,12 @@ import { isEqual } from 'lodash';
 import {
     applyYAxis,
     buildTreeStateFromModel,
+    buildLegendColors,
     ColorAllocator,
     computeYRange,
     getTimeForX as timeForX,
     getXForTime as xForTime,
+    registerSeriesNames,
     zoomRange,
     panRange,
     setSpinnerVisible
@@ -66,6 +68,7 @@ export type AbstractXYOutputState = AbstractTreeOutputState & {
     columns: ColumnHeader[];
     allMax: number;
     allMin: number;
+    legendColors: Record<number, string>;
     cursor?: string;
 };
 
@@ -156,6 +159,7 @@ export abstract class AbstractXYOutputComponent<
     private _debouncedUpdateXY = debounce(() => this.updateXY(), 500);
 
     private colors = new ColorAllocator();
+    private seriesNameById = new Map<number, string>();
 
     constructor(props: P) {
         super(props);
@@ -310,7 +314,8 @@ export abstract class AbstractXYOutputComponent<
                         defaultOrderedIds: built.defaultOrderedIds,
                         collapsedNodes: built.collapsedNodes,
                         checkedSeries: built.checkedSeries,
-                        columns: built.columns as any
+                        columns: built.columns as any,
+                        legendColors: {}
                     },
                     () => {
                         this.updateXY();
@@ -349,6 +354,7 @@ export abstract class AbstractXYOutputComponent<
                     onOrderChange={this.onOrderChange}
                     onOrderReset={this.onOrderReset}
                     headers={this.state.columns}
+                    legendColors={this.state.legendColors}
                 />
             </div>
         ) : undefined;
@@ -464,6 +470,7 @@ export abstract class AbstractXYOutputComponent<
     }
 
     private buildScatterData(seriesObj: XYSeries[]) {
+        registerSeriesNames(this.seriesNameById, seriesObj);
         const dataSetArray = new Array<any>();
         let xValues: bigint[] = [];
         const offset = this.props.viewRange.getOffset() ?? BigInt(0);
@@ -498,13 +505,17 @@ export abstract class AbstractXYOutputComponent<
 
         // flushSync: force immediate state update instead of waiting for React 18's automatic batching
         flushSync(() => {
-            this.setState({ xyData: scatterData });
+            this.setState({
+                xyData: scatterData,
+                legendColors: buildLegendColors(this.state.xyTree, this.colors, this.seriesNameById)
+            });
         });
 
         this.calculateYRange();
     }
 
     private buildXYData(seriesObj: XYSeries[]) {
+        registerSeriesNames(this.seriesNameById, seriesObj);
         const dataSetArray = new Array<any>();
         let xValues: bigint[] = [];
         seriesObj.forEach(series => {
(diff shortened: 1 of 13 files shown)


MatthewKhouzam:
Hey, I like this patch a lot, it is something I was planning on doing when I had breathing time. Thanks, the only issue is that it does not take the colors given by the tsp. therefore the totals are no longer red. Can you fix that? 
Hint from Matthew:
I think you rebuilt the legend instead of using the one supplied first.

Hint from claude:
 PR 404 doesn't reuse the colors the chart already assigned to each line — it runs a second, independently-keyed color allocation for the legend, and at the moment it runs the keys don't line up with the chart's keys. So the swatches walk the palette on their own instead of mirroring the existing line colors.
  
  Here's the mechanism, from the cherry-picked code:
  
  Chart lines get their color in buildXYData (abstract-xy-output-component.tsx):
```  
  const color = this.getSeriesColor(series.seriesName); // -> this.colors.get(series.seriesName)
  ```
  So the ColorAllocator key for a line is the full series.seriesName.
  
  The legend gets its color in buildLegendColors (xy-shared.ts):
  ```
  const key = seriesNameById.get(entry.id) ?? entry.labels?.[0] ?? String(entry.id);
  legendColors[entry.id] = colors.get(key);
  ```
  ColorAllocator.get(key) caches by the string key and hands out palette slots in first-seen order:
  ```
  get(key: string): string {
      let idx = this.map.get(key);
      if (idx === undefined) { idx = this.i % this.palette.length; this.map.set(key, idx); this.i++; }
      return this.palette[idx];
  }
  ```
  The problem is timing plus key identity:
  
  1. buildLegendColors is first called in fetchTree, right when the tree is built:
  ```
     legendColors: buildLegendColors(built.xyTree, this.colors, this.seriesNameById)
  ```
     At that point this.seriesNameById is still empty — it's only populated later inside buildXYData/buildScatterData via
  registerSeriesNames, which run during updateXY() (called after this setState).
  
  2. With an empty seriesNameById, the legend key falls back to entry.labels?.[0] (the leaf display name) for every entry — and it does
  this for all leaf entries in the tree, not just the plotted ones.
  3. When the chart later colors its lines with colors.get(series.seriesName), the key is the qualified series name, which is frequently
  not identical to the tree leaf's labels[0]. Different key → different palette slot. And because the legend already consumed palette
  slots (advancing this.i) for every leaf, the plotted series land on shifted indices.
  
  Net effect: the legend swatches are derived from a fresh, differently-keyed palette walk rather than from the color the chart actually
  drew, so they "ignore" the existing line colors. They'd only coincidentally match when labels[0] === seriesName for every series.
  
  The correct fix is to make the legend use the same identity the chart uses — either look up the color already assigned to each series
  (by seriesId, from the chart datasets) instead of re-deriving it, or at minimum defer/rebuild legendColors after registerSeriesNames
  has populated seriesNameById and key it identically to getSeriesColor. Want me to implement that?

**Comment by hoathaitran3209** [comment]
Hi @MatthewKhouzam, about your comment, I updated it. 
Thank you so much.

**Comment by MatthewKhouzam** [comment]
Almost there! this looks great! You need to enable the legend for the "total" or default fields.

**Comment by hoathaitran3209** [comment]
> Almost there! this looks great! You need to enable the legend for the "total" or default fields.

Done. I also updated the code Minor "Clear seriesNameById before re-registering to avoid stale legend colors for unchecked series."

**Review by MatthewKhouzam — DISMISSED** [review]
Works here! thanks!

**Comment by MatthewKhouzam** [comment]
Fix the issues in linting please also you want to squash the commits and update the commit message? if you used coding assistants such as claude opus, please mention it in the commit message.

**Comment by hoathaitran3209** [comment]
> Fix the issues in linting please also you want to squash the commits and update the commit message? if you used coding assistants such as claude opus, please mention it in the commit message.

Done, I fixed the format:check for the code and also squashed the commit.

**Comment by jonahgraham** [comment]
> update the commit message

@hoathaitran3209 this part is still outstanding. The current commit message is:

```
IDE-92565-add-aengends-colummn

Fix legend rendering for XY charts

Fix legend rendering for XY charts

Apply Prettier formatting
```

Something minimal might be:

```
Add legends column to XY trace viewer
```

A more descriptive commit message may be (remember to suitably word wrap the message)

```
Add legends column to XY trace viewer

This commit adds a legends column to the XY
trace viewer to improve readability and make it easier
to identify plotted series.

Changes
- Add a legends column to the filter tree.
- Update XY output components to support the new column.
- Adjust table layout and styles.
- Refactor related utility code where necessary.
```

jonahgraham:
The "[3PP License Check / 3PP License Check using dash-licenses (ubuntu-latest) (pull_request)](https://github.com/eclipse-cdt-cloud/vscode-trace-extension/actions/runs/29138657907/job/86844914135?pr=404)Failing after 48s" check that is failing is not related to this PR as main. I have raised https://github.com/eclipse-cdt-cloud/vscode-trace-extension/issues/405 to track the issue independently of this PR.

MatthewKhouzam:
> > update the commit message
> 
> @hoathaitran3209 this part is still outstanding. The current commit message is:
> 
> ```
> IDE-92565-add-aengends-colummn
> 
> Fix legend rendering for XY charts
> 
> Fix legend rendering for XY charts
> 
> Apply Prettier formatting
> ```
> 
> Something minimal might be:
> 
> ```
> Add legends column to XY trace viewer
> ```
> 
> A more descriptive commit message may be (remember to suitably word wrap the message)
> 
> ```
> Add legends column to XY trace viewer
> 
> This commit adds a legends column to the XY
> trace viewer to improve readability and make it easier
> to identify plotted series.
> 
> Changes
> - Add a legends column to the filter tree.
> - Update XY output components to support the new column.
> - Adjust table layout and styles.
> - Refactor related utility code where necessary.
> ```

More details. your PR message is great. Your commit message is the item needing love, and that's what stays in our history. But the code is perfect as-is, IMO.

hoathaitran3209:
Hi @MatthewKhouzam and @jonahgraham -san,
I updated the commit message for this PR. 
Thank you for the review.

MatthewKhouzam (approved):

MatthewKhouzam:
@arfio Any thoughts? I think it's good to go!

arfio (approved):
Looks good! Thanks for adding this feature!

